PR #78, opened 2020-10-07
Feature/pmf 23 test destination connection

akhtyamovrr:


adapters/aws_redshift.go
```diff
@@ -25,6 +25,11 @@ type AwsRedshift struct {
 	s3Config        *S3Config
 }
 
+type RedshiftConfig struct {
```
xtreding, line 28: Move this config to connection handler (as Snowflake)

adapters/aws_redshift.go
```diff
@@ -67,6 +72,11 @@ func (ar *AwsRedshift) CreateDbSchema(dbSchemaName string) error {
 	return createDbSchemaInTransaction(ar.dataSourceProxy.ctx, wrappedTx, dbSchemaName)
 }
 
+func (ar *AwsRedshift) Test() error {
+	_, err := ar.dataSourceProxy.dataSource.Query("SELECT 1;")
```
xtreding, line 76: You can use dataSource.Ping() for connection validation

xtreding, line 76: You can use dataSource.Ping() for connection validation

handlers/connection.go
```diff
@@ -0,0 +1,168 @@
...
+		if err = json.Unmarshal(body, &s3Config); err != nil {
+			return err
+		}
+		if err = s3Config.Validate(); err != nil {
+			return err
+		}
+		s3, err := adapters.NewS3(&s3Config)
+		defer s3.Close()
+		return err
+	case "snowflake":
```
xtreding, line 126: Since we've supported only postgres,clickhouse and redshift, we can omit snowflake and s3 test connections, because we don't know config structure (and snowflake can be configured via s3,gcp)

adapters/aws_redshift.go
xtreding: and for other datasources which are sql.DB

xtreding (changes requested):

adapters/aws_redshift.go
xtreding, line 28: do we need RedshiftConfig here?

xtreding (approved):

adapters/aws_redshift.go
akhtyamovrr: I use it as a DTO and use RedshiftConfig at another client that sends configuration for validation

PR #91, opened 2020-10-21
Feature/pmf 32 strict mapping

akhtyamovrr:


storages/factory.go
```diff
@@ -89,7 +90,7 @@ func Create(ctx context.Context, name, logEventPath string, destination Destinat
 		return nil, nil, fmt.Errorf("Unknown destination mode: %s. Available mode: [%s, %s]", destination.Mode, BatchMode, StreamMode)
 	}
 
-	processor, err := schema.NewProcessor(tableName, mapping)
+	processor, err := schema.NewProcessor(tableName, mapping, destination.DataLayout.MappingType)
```
xtreding, line 93: potential panic. destination.DataLayout can be nil.

xtreding (approved):

storages/factory.go
akhtyamovrr: Thnx for noticing! Fixed

PR #124, opened 2020-11-23
Feature/119 sql debug logs

akhtyamovrr:


xtreding:
I think it would be better if Adapters don't know about destination id. I suggest the following structure:

-Adapters have only certain QueryLogger 
--QueryLogger (per destination). It has a destination id and writes it as a prefix.
---DateTimeWriterProxy
----Single file writer/stdout writer depends on the configuration.
In the result,  we have N loggers with 1 output point (file writer of stdout writer).

Also, we should log BigQuery statements. It might be done by writing custom logs like:
"creating table: %s", json.Marshal(bigquery.Schema)

appconfig/appconfig.go
```diff
@@ -35,16 +36,13 @@ func setDefaultParams() {
 	viper.SetDefault("log.show_in_server", false)
 	viper.SetDefault("log.rotation_min", "5")
 	viper.SetDefault("synchronization_service.connection_timeout_seconds", 20)
+	viper.SetDefault("query_log.rotation_min", "5")
```
xtreding, line 39: query_log isn't clear enough. It would be better to use: sql_log or sql_debug_log

logging/query_logger.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	if l.logger != nil {
+		l.logger.Printf("[%s] %s values: [%s]\n", destinationId, query, strings.Join(stringValues, ", "))
+	} else {
+		Debugf("[%s] %s values: [%s]", destinationId, query, strings.Join(stringValues, ", "))
+	}
+}
+
+func (l *QueryLogger) Close() {
+	//l.logger.
```
xtreding, line 56: fix closing

main.go
```diff
@@ -160,7 +160,25 @@ func main() {
 	logEventPath := viper.GetString("log.path")
 
 	//Create event destinations
-	destinationsService, err := destinations.NewService(ctx, destinationsViper, destinationsStr, logEventPath, syncService, storages.Create)
+	var queryLoggerConfig *logging.Config
+	queryLogsEnabled := false
+	if viper.IsSet("query_log.path") {
```
xtreding, line 165: we could do it in the appconfig (there is similar code - global logger initialization)

xtreding (approved):

logging/query_logger.go
akhtyamovrr: fixed

appconfig/appconfig.go
akhtyamovrr: fixed

main.go
akhtyamovrr: fixed

adapters/clickhouse.go
```diff
@@ -190,11 +190,17 @@ type ClickHouse struct {
...
 	nullableFields        map[string]bool
+	queryLogger           *logging.QueryLogger
 }
 
 //NewClickHouse return configured ClickHouse adapter instance
 func NewClickHouse(ctx context.Context, connectionString, database, cluster string, tlsConfig map[string]string,
-	tableStatementFactory *TableStatementFactory, nullableFields map[string]bool) (*ClickHouse, error) {
+	tableStatementFactory *TableStatementFactory, nullableFields map[string]bool,
+	queryLogger *logging.QueryLogger) (*ClickHouse, error) {
+	if queryLogger == nil {
```
xtreding, line 200: It would be better to create test instances with an empty logger and assume that queryLogger parameter can't be nil

adapters/postgres.go
```diff
@@ -100,13 +100,17 @@ func (dsc *DataSourceConfig) Validate() error {
...
+	ctx         context.Context
+	config      *DataSourceConfig
+	dataSource  *sql.DB
+	queryLogger *logging.QueryLogger
 }
 
 //NewPostgres return configured Postgres adapter instance
-func NewPostgres(ctx context.Context, config *DataSourceConfig) (*Postgres, error) {
+func NewPostgres(ctx context.Context, config *DataSourceConfig, queryLogger *logging.QueryLogger) (*Postgres, error) {
+	if queryLogger == nil {
```
xtreding, line 111: same

adapters/snowflake.go
```diff
@@ -86,14 +86,19 @@ func (sc *SnowflakeConfig) Validate() error {
...
+	s3Config    *S3Config
+	dataSource  *sql.DB
+	queryLogger *logging.QueryLogger
 }
 
 //NewSnowflake return configured Snowflake adapter instance
-func NewSnowflake(ctx context.Context, config *SnowflakeConfig, s3Config *S3Config) (*Snowflake, error) {
+func NewSnowflake(ctx context.Context, config *SnowflakeConfig, s3Config *S3Config,
+	queryLogger *logging.QueryLogger) (*Snowflake, error) {
+	if queryLogger == nil {
```
xtreding, line 99: same

destinations/service_test.go
```diff
@@ -108,7 +109,7 @@ func TestServiceInit(t *testing.T) {
 	payload := &payloadHolder{payload: []byte(initialDestinations)}
 	mockDestinationsServer := startTestServer(payload)
 
-	service, err := NewService(context.Background(), nil, mockDestinationsServer.URL, "/tmp", "/tmp/fallback", 5, nil, createTestStorage)
+	service, err := NewService(context.Background(), nil, mockDestinationsServer.URL, "/tmp", "/tmp/fallback", 5, nil, nil, createTestStorage)
```
xtreding, line 112: it would be better to use here empty logger instead of nil

xtreding: and in other places

appconfig/appconfig.go
```diff
@@ -92,6 +119,26 @@ func Init() error {
 	return nil
 }
 
+func InitQueryWriter(globalLogsWriter io.Writer) (io.Writer, error) {
```
xtreding, line 122: I usually use Init() method if the method doesn't return the created value, but sets it to global variable (like log).
If a method returns the created value - it's name is New..()

storages/proxy.go
```diff
@@ -12,10 +12,11 @@ type RetryableProxy struct {
...
 
-	config  *Config
-	storage events.Storage
-	ready   bool
-	closed  bool
+	config      *Config
+	storage     events.Storage
+	ready       bool
+	closed      bool
+	queryLogger *logging.QueryLogger
```
xtreding, line 19: Do we need a query logger here?

appconfig/appconfig.go
```diff
@@ -17,13 +19,15 @@ type AppConfig struct {
 	UaResolver  useragent.Resolver
 
 	AuthorizationService *authorization.Service
+	QueryLogsWriter      *io.Writer
```
xtreding, line 22: io.Writer interface -> it's already a pointer

appconfig/appconfig.go
```diff
@@ -68,6 +89,12 @@ func Init() error {
 	var appConfig AppConfig
 	appConfig.ServerName = serverName
 
+	queryLogsWriter, err := InitQueryWriter(globalLogsWriter)
+	if err != nil {
+		return err
+	}
+	appConfig.QueryLogsWriter = &queryLogsWriter
```
xtreding, line 96: queryLogsWriter is io.Writer (interface). Interfaces are pointers by default.

appconfig/appconfig.go
```diff
@@ -92,6 +119,26 @@ func Init() error {
 	return nil
 }
 
+func InitQueryWriter(globalLogsWriter io.Writer) (io.Writer, error) {
+	var queryLogsWriter io.Writer
+	if viper.IsSet("sql_debug_log.path") {
+		if viper.GetString("sql_debug_log.path") != "global" {
+			queryLoggerConfig := logging.Config{
+				LoggerName:  "query",
```
xtreding, line 127: In this case, log files will have names like hostname-query-yyyy-mm... 
"query" isn't clear enough in this case. It would be better to use sql-debug or something like that.

xtreding (approved):

destinations/service_test.go
akhtyamovrr: It's not queryLogger here, it's io.Writer. In fact, it may be nil if SQL debug logs are disabled. So, I think that here it may be nil.

storages/proxy.go
akhtyamovrr: Fixed

adapters/snowflake.go
akhtyamovrr: fixed

adapters/clickhouse.go
akhtyamovrr: fxied

adapters/postgres.go
akhtyamovrr: fixed

appconfig/appconfig.go
akhtyamovrr: fixed

appconfig/appconfig.go
akhtyamovrr: fixed

appconfig/appconfig.go
akhtyamovrr: Fixed

appconfig/appconfig.go
akhtyamovrr: Fixed

destinations/service_test.go
akhtyamovrr: I've fixed queryLogger nil -> &QueryLogger{} at other places at tests

PR #133, opened 2020-12-01
some fixes

xtreding:


xtreding:
https://github.com/jitsucom/eventnative/pull/133/files#diff-fa1be9fa613b8f1a391e8dd2c572ae5a899efebfe8450b7fc592d818d429eb74R121
Would it be better not to have `cast` action, but if `type` present in mapping object than field value will be cast?

design-docs/pipeline.md
```diff
@@ -124,7 +124,7 @@ mapping:
 Following field actions are supported:
 
 * **move** — move JSON subtree to another node. 
-* **remove** - remove JSON subtree (dst param  is not needed)
+* **remove** - remove JSON subtree (dst param is not needed)
 * **cast** – assign an explicit type to a node (dst param is not needed)
 
 Actions can be chained:
```
xtreding, line 130: we can delete this topic if we didn't have action `cast`. But apply casting if `type` field is specified in field object

PR #149, opened 2020-12-22
Feature/bulk update

akhtyamovrr:


sources/sync_task.go
```diff
@@ -93,13 +92,13 @@ func (st *SyncTask) Sync() {
...
-			events.EnrichWithEventId(object, getHash(object))
+			events.EnrichWithEventId(object, uuid.GetHash(object))
 			events.EnrichWithCollection(object, st.collection)
 			events.EnrichWithTimeInterval(object, intervalToSync)
 		}
-
 		for _, storage := range st.destinations {
-			rowsCount, err := storage.SyncStore(objects)
+			rowsCount, err := storage.SyncStore(collectionTable, objects, intervalToSync.String())
+			//rowsCount, err := storage.SyncStore(collectionTable, objects, штеук)
```
xtreding, line 101: del

events/storage.go
```diff
@@ -8,7 +8,7 @@ type Storage interface {
 	io.Closer
 	Store(fileName string, payload []byte, alreadyUploadedTables map[string]bool) (map[string]*StoreResult, int, error)
 	StoreWithParseFunc(fileName string, payload []byte, skipTables map[string]bool, parseFunc func([]byte) (map[string]interface{}, error)) (map[string]*StoreResult, int, error)
-	SyncStore([]map[string]interface{}) (int, error)
+	SyncStore(tableName string, objects []map[string]interface{}, deleteValue string) (int, error)
```
xtreding, line 11: rename deleteValue -> timeIntervalValue ?

adapters/postgres.go
```diff
@@ -414,8 +414,52 @@ func (p *Postgres) Insert(table *Table, valuesMap map[string]interface{}) error
...
+	}
+	if err := p.insertInTransaction(wrappedTx, table, objects); err != nil {
+		wrappedTx.Rollback()
+		return err
+	}
+	return wrappedTx.DirectCommit()
+}
+
+func (p *Postgres) deleteInTransaction(wrappedTx *Transaction, table *Table, deleteConditions *DeleteConditions) error {
+	deleteQueryTemplate := "DELETE FROM %s.%s WHERE %s"
```
xtreding, line 434: move to constants

adapters/types.go
```diff
@@ -0,0 +1,36 @@
+package adapters
```
xtreding, line 1: It would be better to rename types.go -> delete_condition.go

adapters/types.go
```diff
@@ -0,0 +1,36 @@
...
+
+type DeleteCondition struct {
+	Field  string
+	Value  interface{}
+	Clause string
+}
+
+func (dc *DeleteCondition) ToQueryString() string {
+	var value string
+	switch dc.Value.(type) {
```
xtreding, line 16: we can't trust the value type because there might be explicit type casts for example:
a field might be `integer` but has explicit type `text` in mappings and in DWH.

xtreding (changes requested):

sources/sync_task.go
akhtyamovrr: fixed

events/storage.go
akhtyamovrr: sure

adapters/postgres.go
akhtyamovrr: fixed

adapters/types.go
akhtyamovrr: changed

adapters/types.go
akhtyamovrr: Delete conditions are converted to queries at adapters now, so, type casts and placeholders are used based on type of destination

adapters/clickhouse.go
```diff
@@ -430,24 +431,69 @@ func (ch *ClickHouse) Insert(table *Table, valuesMap map[string]interface{}) err
...
+	}
+	return wrappedTx.DirectCommit()
+}
+
+func (ch *ClickHouse) deleteInTransaction(wrappedTx *Transaction, table *Table, deleteConditions *DeleteConditions) error {
+	deleteCondition, values := ch.toDeleteQuery(deleteConditions)
+	deleteQuery := fmt.Sprintf(deleteQueryChTemplate, ch.database, table.Name, deleteCondition)
+	deleteStmt, err := wrappedTx.tx.PrepareContext(ch.ctx, deleteQuery)
+	if err != nil {
+		wrappedTx.Rollback()
```
xtreding, line 455: don't rollback transaction in "*inTransaction" methods. It will happen in an upper level

xtreding (approved):

adapters/clickhouse.go
akhtyamovrr: 👍

PR #176, opened 2021-01-18
Feature/redis driver

akhtyamovrr:


drivers/factory.go
```diff
@@ -97,7 +97,8 @@ func Create(ctx context.Context, name string, sourceConfig *SourceConfig) (map[s
 	for _, collection := range collections {
 		driver, err := createDriverFunc(ctx, sourceConfig, collection)
 		if err != nil {
-			return nil, fmt.Errorf("error creating [%s] driver for [%s] collection: %v", sourceConfig.Type, collection, err)
+			logging.Errorf("error creating [%s] driver for [%s] collection: %v", sourceConfig.Type, collection, err)
```
akhtyamovrr, line 100: I suppose it's better to skip the wrong collection configurations that not create any instances at all. If someone misconfigured one collection, others should not stop syncing. But logs message would be enough to troubleshoot why the wrong collection is not syncronized

xtreding (approved):

PR #183, opened 2021-01-25
Ga improvement

akhtyamovrr:


adapters/postgres.go
```diff
@@ -479,48 +485,77 @@ func (p *Postgres) BulkInsert(table *Table, objects []map[string]interface{}) er
...
 		}
 	}
-
+	if len(valueArgs) > 0 {
+		err := p.executeInsert(wrappedTx, table, header, placeholdersBuilder, valueArgs)
+		if err != nil {
+			return err
+		}
+	}
+	logging.Infof("Inserted [%d] rows in [%.2f] seconds", len(objects), time.Now().Sub(start).Seconds())
```
xtreding, line 548: This log message isn't clear. We should know the context (destination id).
Maybe we could move it to SQL debug logs?

adapters/postgres.go
```diff
@@ -479,48 +485,77 @@ func (p *Postgres) BulkInsert(table *Table, objects []map[string]interface{}) er
...
 	for _, row := range objects {
-		var values []interface{}
-		for _, column := range header {
+		// if number of values exceeds limit, we have to execute insert query on processed rows
+		if len(valueArgs)+len(header) > postgresValuesLimit {
+			err := p.executeInsert(wrappedTx, table, header, placeholdersBuilder, valueArgs)
+			if err != nil {
+				return err
+			}
+			logging.Infof("Inserted %d rows", len(valueArgs)/len(table.Columns))
```
xtreding, line 507: This log message is useless

adapters/postgres.go
```diff
@@ -479,48 +485,77 @@ func (p *Postgres) BulkInsert(table *Table, objects []map[string]interface{}) er
 }
 
 func (p *Postgres) insertInTransaction(wrappedTx *Transaction, table *Table, objects []map[string]interface{}) error {
```
xtreding, line 487: It would be better to rename this func to batchInsertInTransaction or bulkInsertInTransaction

xtreding (approved):

adapters/postgres.go
akhtyamovrr: I'm not sure if the message should be at SQL logs. SQL logs idea was that we can debug queries. This information may be useful in application logs (to see the time spent to insert data in case something goes wrong or slow). So, I suppose that this log should be in app logs. I think it may be in Debug mode.

adapters/postgres.go
akhtyamovrr: +, it was there just to check how many objects are inserted in one batch

adapters/postgres.go
akhtyamovrr: ✅

adapters/postgres.go
akhtyamovrr: Also, at the adapter level, we don't know about the destination id. We may pass it as an additional parameter when an adapter is created. Or we may log data from data source configuration (host, schema, database + table name). The second option seems more confusing, so, I suppose that it's better to let the adapter know the destination id adapter belongs to.

akhtyamovrr: Moved logging to the storage level to keep destination id

PR #209, opened 2021-03-01
Fix creating empty files at log closing

Alessar:
There is a possibility that empty log files will be created at the closing log fabric.
The fix allows to suppress such empty log files creating.


xtreding:
@Alessar Could you please change this PR to the `beta` branch?

PR #218, opened 2021-03-04
Parse ISO time in strings

### LINKED ISSUE #171: Parse json strings with datetime as SQL timestamp
### Problem

At present *EventNative* parse all JSON strings as strings and creates SQL columns with text type by default (if typecast isn't configured). Sometimes it is inconvenient to configure explicit SQL timestamp typecasts to every field. It can be recognized by default.

### Solution

All input JSON strings with ISO DateTime format patterns (`2006-01-02T15:04:05.000000Z` or `2006-01-02T15:04:05.000Z`) must be parsed as timestamp and be mapped into SQL type as SQL timestamp.

Alessar:
Adding capability to parse ISO time from string and convert it into time.Time value.
Fixes #171

### DIFF
--- a/typing/datatype.go
+++ b/typing/datatype.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/jitsucom/eventnative/logging"
-	"github.com/jitsucom/eventnative/timestamp"
 )
 
 type DataType int
@@ -110,7 +109,7 @@ func ReformatTimeValue(value interface{}) interface{} {
 		return value
 	}
 
-	timeValue, err := time.Parse(timestamp.Layout, stringValue)
+	timeValue, err := time.Parse(time.RFC3339Nano, stringValue)
 	if err == nil {
 		return timeValue
 	}
--- a/typing/datatype.go
+++ b/typing/datatype.go
@@ -3,9 +3,11 @@ package typing
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/jitsucom/eventnative/logging"
 	"strings"
 	"time"
+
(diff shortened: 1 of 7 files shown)


typing/datatype.go
```diff
@@ -101,6 +103,21 @@ func ReformatValue(v interface{}) interface{} {
...
 }
 
+// ReformatTimeValue processes string with ISO DateTime into time.Time
+func ReformatTimeValue(value interface{}) interface{} {
+	stringValue, ok := value.(string)
+	if !ok {
+		return value
+	}
+
+	timeValue, err := time.Parse(timestamp.Layout, stringValue)
```
xtreding, line 113: Since we're going to support both formats (`timestamp.Layout` and `timestamp.DeprecatedLayout`),
please add parsing `timestamp.DeprecatedLayout` as well. (consider rename DeprecatedLayout into something without `Deprecated`)

xtreding (changes requested):

typing/datatype.go
Alessar: My suggestion is to remove `DeprecatedLayout` at all.
Formatting timestamp is better for using only one `Layout` format.
Parsing is better to use from all format as `RFC3339Nano` - feel free to see `TestReformatTimeValue`

xtreding: @Alessar  It is a good idea. Then it would be better to refactor func `stringToTimestamp()` in `typing.Converter` - replace logic with `Layout` and `DeprecatedLayout` with your solution. Then `DeprecatedLayout` can be deleted.

Issue #123, opened 2020-11-22
Error during non-docker app start


vklimontovich:
## Description

If the app built from sources and https://docs.eventnative.org/deployment/build-from-sources and started without config, error message appears in log

## Steps to reproduce / actual behavior

Build binary from sources and run it without config. See following error in logs

```
[ERROR]: open /home/eventnative/app/res/: no such file or directory
```

## Expected behavior

Error shouldn't happen. All defaults config variables should be rather paths relative to binary, not a values convenient for docker environment. The app should create those folders and put clear message of the location in output. Example:
 - Log directory ~/go/src/github.com/jitsucom/eventnative/build/dist/logs is missing, creating
 - Resources directory ~/go/src/github.com/jitsucom/eventnative/build/dist/logs is missing, creating

## Notes

To keep docker build functioning we need to:
 - Supply default ./eventnative.docker.yaml with /home/eventnative/* as default path
 - Support config merging

PR #221, opened 2021-03-06
Create logs directory and fix app starting issues

Alessar:
Fix issue: #123
* `-cfg` parameter becomes optional
* fixing issues during startup without config file
* Creating `log.path` directory for log files
* Correcting error message that maxmind db is missing
* Writing info about `server.log.path`
* Different default parameters for run with and without container
* Fix `FatalF` implementation

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,10 +3,8 @@ package main
 import (
 	"bytes"
 	"context"
-	"errors"
 	"flag"
 	"fmt"
-	"github.com/jitsucom/eventnative/scheduling"
 	"math/rand"
 	"net/http"
 	"os"
@@ -37,6 +35,7 @@ import (
 	"github.com/jitsucom/eventnative/resources"
 	"github.com/jitsucom/eventnative/routers"
 	"github.com/jitsucom/eventnative/safego"
+	"github.com/jitsucom/eventnative/scheduling"
 	"github.com/jitsucom/eventnative/singer"
 	"github.com/jitsucom/eventnative/sources"
 	"github.com/jitsucom/eventnative/storages"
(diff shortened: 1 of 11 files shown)


xtreding:
There should be:
1) different viper default configuration parameters for the containerized run and binary run different should be applied depends on `cr` flag. (see main.go `containerizedRun`)


|config var                                      | containerized Run                       |  run from sources |
|----------                                      | ---                                                 | ---                          |
|`server.log.path`                          | `/home/eventnative/logs`          | `./logs`                  |
|`log.path`                                     | `/home/eventnative/logs/events` | `./logs/events`   |
|`geo.maxmind_path`                  |   `/home/eventnative/app/res/`  | `./`                         |

Consider adding log messages for example `log.path directory is missing, creating` for `log.path` and `server.log.path`

xtreding (changes requested):
Added global comment

main.go
```diff
@@ -86,36 +86,34 @@ func readInViperConfig() error {
...
 
-	err = viper.ReadConfig(bytes.NewBuffer(payload.Content))
-	if err != nil {
-		errWithContext := fmt.Errorf("Error reading/parsing viper config from %s: %v", configSourceStr, err)
-		return handleConfigErr(errWithContext)
+	if payload != nil {
+		err = viper.ReadConfig(bytes.NewBuffer(payload.Content))
+		if err != nil {
+			errWithContext := fmt.Errorf("Error reading/parsing viper config from %s: %v", configSourceStr, err)
+			return handleConfigErr(errWithContext)
```
vklimontovich, line 115: I would probably refrain from using the internal libs name ("viper" in this case) at all in error messages. The messages will be confusing for most people.

"Error reading/parsing config file %s: %v" would make more sense

Alessar: Fixed

Alessar:
> Consider adding log messages for example `log.path directory is missing, creating` for `log.path` and `server.log.path`

Current behaviour:
* Creating directory: `log.path`
* Showing directory `server.log.path`

`geo.maxmind_path` - looks like it is the path for reading, not writing, so it is not necessary to create it.

xtreding (approved):

PR #235, opened 2021-03-14
158. Validate stage permissions on test connection

### LINKED ISSUE #158: TestConnection must validate stage permissions
## Description
Stage - intermediate layer in batch mode integrations e.g. Google Cloud Storage, s3, Snowflake Stage. At present **EventNative** destinations test connection functionality doesn't validate stage permissions, but it should. As a result destination integration won't work, but the test connection is ok.

## Steps to reproduce

call `/destinations/test` with a destination in batch mode (without stage write permissions).

## Expected behavior

The result must contain an error with a description about permissions.

## Actual behavior

Test connection is ok

Alessar:
Fixes #158
* Adding function for validating WRITE permissions on S3
* Adding permission S3 validation on batch mode at TestConnection to RedShift
* Adding permission S3 validation on batch mode at TestConnection to Snowflake
* Adding function for validating WRITE permissions on Google Cloud Storage
* Adding permission GCS validation on batch mode at TestConnection to BigQuery
* Adding permission GCS validation on batch mode at TestConnection to Snowflake

### DIFF
--- a/adapters/google_cloud_storage.go
+++ b/adapters/google_cloud_storage.go
@@ -1,15 +1,18 @@
 package adapters
 
 import (
-	"cloud.google.com/go/storage"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"google.golang.org/api/iterator"
-	"google.golang.org/api/option"
 	"io/ioutil"
 	"strings"
+
+	"cloud.google.com/go/storage"
+	"github.com/jitsucom/eventnative/logging"
+	"github.com/jitsucom/eventnative/timestamp"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
 )
 
 type GoogleCloudStorage struct {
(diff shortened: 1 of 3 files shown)


adapters/google_cloud_storage.go
```diff
@@ -136,6 +139,24 @@ func (gcs *GoogleCloudStorage) GetObject(key string) ([]byte, error) {
 	return ioutil.ReadAll(r)
 }
 
+// Function tries to create temporary file and remove it.
+// Returns OK if file creation was successfull.
```
xtreding, line 143: Let's write comments to functions in single format like:
```
//NameOfFunction tries to create ....
//returns nil if ok and error ....
```

Because this function doesn't return OK, it returns nil or error. Comments style is described [here](https://blog.golang.org/godoc).

xtreding (approved):

PR #241, opened 2021-03-20
140. Adding column count limit

### LINKED ISSUE #140: Limit number of columns
### Problem

Currently, a mistake in the code can generate an endless number of columns in the database.

### Solution

Let's have a setting `max_columns` (per destination) that is 100 by default. If during schema patching number of columns has exceeded, let's fail the whole process - row or whole batch.

Alessar:
Fixes #140 
* Adding server.max_columns setting
* Adding column count calculation
* Adding preventing increasing columns in databases
* Adding data_layout.max_columns setting
* Adding overriding behaviour for max_columns settings.


server/storages/bigquery.go
```diff
@@ -71,7 +71,7 @@ func NewBigQuery(config *Config) (Storage, error) {
 		return nil, err
 	}
 
-	tableHelper := NewTableHelper(bigQueryAdapter, config.monitorKeeper, config.pkFields, adapters.SchemaToBigQueryString, config.streamMode)
+	tableHelper := NewTableHelper(bigQueryAdapter, config.monitorKeeper, config.pkFields, adapters.SchemaToBigQueryString, config.streamMode, config.max_columns)
```
vklimontovich, line 75: Shouldn't it be maxColumns instead of max_columns?

server/config/config.template.yaml
```diff
@@ -40,6 +40,11 @@ server:
   ### Public URL. It is used in welcome.html if not configured it will be taken from 'Host' http header on welcome.html requests
   #public_url: https://yourhost #Optional.
 
+  ### Limit the count of columns for any destinations.
+  ### It helps to prevent generating an infinite amount of columns in databases.
+  ### Zero value turns off column count calculation.
+  max_columns: 100 #Optional
```
xtreding, line 46: I think it would be better to have this flag on the destination configuration level (in `data_layout` section). 
```
destinations:
  pg:
   ...
    data_layout:
      max_columns: 100
```
The logic might be:
- if a destination doesn't have configured `max_columns` - apply default value = `100`

server/storages/bigquery.go
xtreding: Absolutely

server/storages/table_helper.go
```diff
@@ -26,11 +26,12 @@ type TableHelper struct {
 	pkFields           map[string]bool
 	columnTypesMapping map[typing.DataType]string
 
-	streamMode bool
+	streamMode  bool
+	max_columns int
```
xtreding, line 30: `max_columns` -> `maxColumns`

server/storages/table_helper.go
```diff
@@ -88,6 +90,14 @@ func (th *TableHelper) EnsureTable(destinationName string, dataSchema *adapters.
 		return dbSchema, nil
 	}
 
+	if th.max_columns > 0 {
+		columns_count := len(dbSchema.Columns) + len(diff.Columns)
+		if columns_count > th.max_columns {
+			msg := fmt.Sprintf("Count of columns %d should be less or equal max_columns %d", columns_count, th.max_columns)
```
xtreding, line 96: return `fmt.Errorf("", ...)` could be used here.

xtreding (changes requested):

server/config/config.template.yaml
Alessar: It was already mentioned at the [issue](https://github.com/jitsucom/jitsu/issues/140#issuecomment-803409720).
It looks like a simple error checking.
Will it be different for different destinations?
If not - it can be used just for simple checking.

server/storages/table_helper.go
Alessar: Renamed

server/storages/table_helper.go
Alessar: Changed

server/storages/bigquery.go
Alessar: Renamed

server/config/config.template.yaml
Alessar: Fixed.
To error checking, an option is used on the server level if it is positive.
Any destination can override that setting by any positive value.

server/config/config.template.yaml
```diff
@@ -40,6 +40,12 @@ server:
   ### Public URL. It is used in welcome.html if not configured it will be taken from 'Host' http header on welcome.html requests
   #public_url: https://yourhost #Optional.
 
+  ### Limit the count of columns for any destinations.
+  ### It helps to prevent generating an infinite amount of columns in databases.
+  ### Zero value turns off the column count calculation.
+  ### It can be overridden at the destination level only by a positive value.
+  max_columns: 100 # Optional
```
xtreding, line 47: We shouldn't have global value for `max_columns`

server/main.go
```diff
@@ -293,7 +293,9 @@ func main() {
 		logging.Warnf("Global users recognition isn't configured")
 	}
 
-	destinationsFactory := storages.NewFactory(ctx, logEventPath, coordinationService, eventsCache, loggerFactory, globalRecognitionConfiguration)
+	maxColumns := viper.GetInt("server.max_columns")
```
xtreding, line 296: We shouldn't have global value for `max_columns`

server/storages/facebook.go
```diff
@@ -39,7 +39,7 @@ func NewFacebook(config *Config) (Storage, error) {
 	requestDebugLogger := config.loggerFactory.CreateSQLQueryLogger(config.name)
 	fbAdapter := adapters.NewFacebookConversion(fbConfig, requestDebugLogger)
 
-	tableHelper := NewTableHelper(fbAdapter, config.monitorKeeper, config.pkFields, adapters.SchemaToFacebookConversion, config.streamMode)
+	tableHelper := NewTableHelper(fbAdapter, config.monitorKeeper, config.pkFields, adapters.SchemaToFacebookConversion, config.streamMode, config.maxColumns)
```
xtreding, line 42: We shouldn't have max_columns for facebook (because facebook doesn't have columns)

server/storages/factory.go
```diff
@@ -145,6 +151,10 @@ func (f *FactoryImpl) Create(name string, destination DestinationConfig) (Storag
 		for _, field := range destination.DataLayout.PrimaryKeyFields {
 			pkFields[field] = true
 		}
+
+		if destination.DataLayout.MaxColumns > 0 {
```
xtreding, line 155: There must be a default value = 100, and the destination can override it.

server/storages/google_analytics.go
```diff
@@ -39,7 +39,7 @@ func NewGoogleAnalytics(config *Config) (Storage, error) {
 	requestDebugLogger := config.loggerFactory.CreateSQLQueryLogger(config.name)
 	gaAdapter := adapters.NewGoogleAnalytics(gaConfig, requestDebugLogger)
 
-	tableHelper := NewTableHelper(gaAdapter, config.monitorKeeper, config.pkFields, adapters.SchemaToGoogleAnalytics, config.streamMode)
+	tableHelper := NewTableHelper(gaAdapter, config.monitorKeeper, config.pkFields, adapters.SchemaToGoogleAnalytics, config.streamMode, config.maxColumns)
```
xtreding, line 42: We shouldn't have max_columns for google analytics (because google analytics doesn't have columns)

xtreding (changes requested):

server/storages/facebook.go
Alessar: Zero was set for facebook destination type

server/storages/google_analytics.go
Alessar: Zero was set for google analitics destination type

server/storages/factory.go
Alessar: Message was added

server/config/config.template.yaml
Alessar: * `100` in the code - it looks like the magic number and it cannot be changed.
* Configuration settings provide a better description and provide the capability to change it.
* There is the capability to make `self-hosted` installation.
It is better to use one single option than several options, if admins want to change such setting.

xtreding (approved):

PR #242, opened 2021-03-22
170. Several user recognition fields

### LINKED ISSUE #170: Users recognition improvements
### Problem

At present **EventNative** updates only 1 configured field (`internal_id` by default) in all previous events. It is configured globally or on the destination level. (see default values in `appconfig.go` It would be better to have an opportunity to configure N fields which will be updated in all previous events e.g. both `internal_id` and `email`.

### Solution

Configuration parameter`user_id_node` will be renamed into e.g. `identification_nodes` with array type and if at least one of such is present - update all previous events with all non-null/non-empty values from the incoming event.
`Note: leave old config parameter as-is for backward compatibility` and write warning log message (something about that it is deprecated).

Alessar:
Fixes #170
`users_recognition` contains `anonymous_id_node` and list of `identification_nodes`
So if the event contains all fields from `identification_nodes` - in that case, all previous events with the same `anonymous_id_node` will be filled by values of the newly received event.


server/users/service.go
```diff
@@ -112,17 +133,19 @@ func (rs *RecognitionService) start() {
 			}
 
 			for destinationID, identifiers := range rp.DestinationsIdentifiers {
-				//recognized
```
xtreding, line 115: run pipeline only if all properties are present

server/users/service.go
```diff
@@ -30,7 +31,7 @@ type RecognitionPayload struct {
 type EventIdentifiers struct {
 	AnonymousID string
 	EventID     string
-	UserID      string
+	Properties  []interface{}
```
xtreding, line 34: Properties -> IdentificationValues

server/users/service.go
```diff
@@ -242,9 +265,10 @@ func (rs *RecognitionService) runPipeline(destinationID string, identifiers Even
 			continue
 		}
 
-		err = configuration.UserIDJSONPath.Set(event, identifiers.UserID)
+		err = configuration.PropertyJSONPathes.Set(event, identifiers.Properties)
```
xtreding, line 268: Using JSON paths array with values array is not reliable. What happens when the order of `identification_nodes` is changed (e.g. by a user or by JSON serializer in the case when the server loads destinations configuration from URL source)? Wrong values will be set and all recognized events will contain an error. Let's think about how to avoid this situation? (e.g. by using hashmap)

xtreding (changes requested):

server/config/config.template.yaml
```diff
@@ -146,7 +146,8 @@ destinations:
 #    users_recognition: #Optional
 #      enabled: true #set false for disabling
 #      anonymous_id_node: /user/anonymous_id #Required if enabled
-#      user_id_node: /user/id #Required if enabled
+#      identification_nodes:
+#        - /user/id # Required if enabled
```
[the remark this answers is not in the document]
Alessar: Fixed

server/config/config.template.yaml
```diff
@@ -363,7 +364,8 @@ destinations:
 #users_recognition:
 #  enabled: true
 #  anonymous_id_node: /eventn_ctx/user/anonymous_id #Optional. Json path to user anonymous id. Default value is  /eventn_ctx/user/anonymous_id
-#  user_id_node: /eventn_ctx/user/email #Optional. Json path to identified user id. Default value is  /eventn_ctx/user/internal_id
+#  identification_nodes:
+#    - /eventn_ctx/user/email # Optional. Json path to identified user id. Default value is /eventn_ctx/user/internal_id
```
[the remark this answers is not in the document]
Alessar: Fixed

server/storages/factory.go
```diff
@@ -68,8 +70,17 @@ func (ur *UsersRecognition) Validate() error {
 			return errors.New("anonymous_id_node is required")
 		}
 
-		if ur.UserIDNode == "" {
-			return errors.New("user_id_node is required")
+		if 0 == len(ur.PropertyNodes) {
```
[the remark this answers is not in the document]
Alessar: Rewritten

server/users/service.go
Alessar: Fixed

server/storages/factory.go
```diff
@@ -53,9 +54,10 @@ type DataLayout struct {
 }
 
 type UsersRecognition struct {
-	Enabled         bool   `mapstructure:"enabled" json:"enabled,omitempty" yaml:"enabled,omitempty"`
-	AnonymousIDNode string `mapstructure:"anonymous_id_node" json:"anonymous_id_node,omitempty" yaml:"anonymous_id_node,omitempty"`
-	UserIDNode      string `mapstructure:"user_id_node" json:"user_id_node,omitempty" yaml:"user_id_node,omitempty"`
+	Enabled         bool     `mapstructure:"enabled" json:"enabled,omitempty" yaml:"enabled,omitempty"`
+	AnonymousIDNode string   `mapstructure:"anonymous_id_node" json:"anonymous_id_node,omitempty" yaml:"anonymous_id_node,omitempty"`
+	PropertyNodes   []string `mapstructure:"identification_nodes" json:"identification_nodes,omitempty" yaml:"identification_nodes,omitempty"`
```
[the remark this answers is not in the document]
Alessar: Renamed

server/users/service.go
Alessar: Renamed

server/users/service.go
Alessar: Rewritten to use map

server/users/service.go
```diff
@@ -254,10 +266,18 @@ func (rs *RecognitionService) runPipeline(destinationID string, identifiers Even
 			continue
 		}
 
-		err = rs.metaStorage.DeleteAnonymousEvent(destinationID, identifiers.AnonymousID, storedEventID)
-		if err != nil {
-			logging.SystemErrorf("[%s] Error deleting stored recognized event [%s]: %v", destinationID, storedEventID, err)
-			continue
+		// If any property is empty event will be anonymous until filling out that last property.
+		// Removing event from storage is suitable only when all properties are filled out.
+		if configuration.IdentificationJSONPathes.IsFullFilled(event) {
```
xtreding, line 271: This condition could be omitted because `runPipeline()` call happens only if All `identification_nodes` are present.
(IsFullFilled() func should be deleted from JSONPathes struct).

You can just keep the old piece of code here.

server/jsonutils/path.go
```diff
@@ -117,3 +117,74 @@ func FormatPrefixSuffix(key string) string {
 	}
 	return key
 }
+
+type JSONPathes struct {
```
xtreding, line 121: Please write some tests on `JSONPathes`

server/jsonutils/path.go
```diff
@@ -117,3 +117,74 @@ func FormatPrefixSuffix(key string) string {
...
+		if result != "" {
+			result += ", "
+		}
+		result += key
+	}
+
+	return "[" + result + "]"
+}
+
+func (jpa *JSONPathes) Get(object map[string]interface{}) (map[string]interface{}, bool) {
```
xtreding, line 150: Please add comments about returning values (it's not clear now).

server/jsonutils/path.go
```diff
@@ -117,3 +117,74 @@ func FormatPrefixSuffix(key string) string {
...
+	for key, path := range jpa.pathes {
+		value, answer := path.Get(object)
+		array[key] = value
+		result = result || answer
+	}
+
+	return array, result
+}
+
+func (jpa *JSONPathes) Set(object map[string]interface{}, values map[string]interface{}) error {
```
xtreding, line 163: same

server/jsonutils/path.go
```diff
@@ -117,3 +117,74 @@ func FormatPrefixSuffix(key string) string {
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func (jpa *JSONPathes) IsFullFilled(object map[string]interface{}) bool {
```
xtreding, line 177: You could delete this func

server/storages/factory.go
```diff
@@ -68,8 +70,13 @@ func (ur *UsersRecognition) Validate() error {
 			return errors.New("anonymous_id_node is required")
 		}
 
-		if ur.UserIDNode == "" {
-			return errors.New("user_id_node is required")
+		if len(ur.IdentificationNodes) == 0 {
+			if ur.UserIDNode == "" {
+				return errors.New("identification_nodes is required")
+			} else {
+				logging.Warn("user_id_node is deprecated. Please use identification_nodes instead")
```
xtreding, line 77: Please add full path to config parameter:

```
logging.Warn("users_recognition.user_id_node is deprecated. Please use users_recognition.identification_nodes instead")
```

server/users/service.go
```diff
@@ -39,6 +40,16 @@ func RecognitionPayloadBuilder() interface{} {
 	return &RecognitionPayload{}
 }
 
+func (ei *EventIdentifiers) IsAllProperties() bool {
```
xtreding, line 43: It would be more clear to have a func `AllIdentificationValuesFilled` or something like that. `Properties` - is general, and isn't clear enough.

server/users/service.go
```diff
@@ -112,17 +123,17 @@ func (rs *RecognitionService) start() {
 			}
 
 			for destinationID, identifiers := range rp.DestinationsIdentifiers {
-				//recognized
-				if identifiers.UserID != "" {
-					err := rs.runPipeline(destinationID, identifiers)
+				if !identifiers.IsAllProperties() {
```
xtreding, line 126: Please try to avoid negative conditions, for instance this statement will be more clear:

```
if identifiers.IsAllProperties() {
  // Run pipeline only if all properties were recognized - it is needed to update all other anonymous events
  err = rs.runPipeline(destinationID, identifiers)
  if err != nil {
    logging.SystemErrorf("[%s] Error running recognizing pipeline: %v", destinationID, err)
  }
					
} else {
  // If some property is missing - event is anonymous
  err = rs.metaStorage.SaveAnonymousEvent(destinationID, identifiers.AnonymousID, identifiers.EventID, string(rp.EventBytes))
  if err != nil {
    logging.SystemErrorf("[%s] Error saving event with anonymous id %s: %v", destinationID, identifiers.AnonymousID, err)
  }
}
```

xtreding (changes requested):

server/users/service.go
Alessar: Condition was removed

server/jsonutils/path.go
Alessar: Comment was added

server/jsonutils/path.go
Alessar: Comment was added

server/storages/factory.go
Alessar: Changed

server/users/service.go
Alessar: Renamed to `IsAllIdentificationValuesFilled`

server/users/service.go
Alessar: Changed

server/jsonutils/path.go
Alessar: Function was removed

server/jsonutils/path.go
Alessar: Tests were added

server/users/service.go
```diff
@@ -211,7 +219,11 @@ func (rs *RecognitionService) runPipeline(destinationID string, identifiers Even
 	}
 
 	for storedEventID, storedSerializedEvent := range eventsMap {
-		event := map[string]interface{}{}
+		if storedEventID == identifiers.EventID {
```
xtreding, line 222: Do we really need this check?

xtreding (changes requested):

server/users/service.go
Alessar: In current behaviour, it can be removed.

PR #324, opened 2021-04-24
236. Validate write permissions for all destinations

### LINKED ISSUE #236: Validate destinations permissions
### Problem

In continuation of #158 **EventNative** should validate not only stage write permissions, but also destination create tables, table write permissions in both `batch` and `stream` modes. It allows users to know about permissions troubles faster.

### Solution

Function `/destinations/test` should contain the following logic (for SQL destinations):
**Stream destination**:
- create table with name `test_$timestamp_$randomNumber`
- insert test object to test table
- drop test table

**Batch destination**:
- create table with name `test_$timestamp_$randomNumber`
- write a file with test object to stage layer
- execute `Copy` from the stage into the table 
- drop the file from stage
- drop test table

Alessar:
Fix #236
* Adding permission validation as creating tables and inserting values for destinations:
Stream mode: `BigQuery`, `ClickHouse`, `Postgres`, `RedShift`, `Snowflake`.
* Adding permission validations as uploading files and copying values to tables for destinations:
Batch mode: `BigQuery`, `RedShift`, `Snowflake`.

### DIFF
--- a/server/adapters/aws_redshift.go
+++ b/server/adapters/aws_redshift.go
@@ -3,12 +3,10 @@ package adapters
 import (
 	"context"
 	"fmt"
-	"math/rand"
 	"strconv"
 	"strings"
 
 	"github.com/jitsucom/jitsu/server/logging"
-	"github.com/jitsucom/jitsu/server/timestamp"
 	"github.com/jitsucom/jitsu/server/typing"
 	_ "github.com/lib/pq"
 )
@@ -314,37 +312,4 @@ func (ar *AwsRedshift) recreateNotNullColumnInTransaction(wrappedTx *Transaction
 //Close underlying sql.DB
 func (ar *AwsRedshift) Close() error {
 	return ar.dataSourceProxy.Close()
-}
-
-func (ar *AwsRedshift) ValidateWritePermission() error {
-	tableName := fmt.Sprintf("test_%v_%v", timestamp.NowUTC(), rand.Int())
-	columnName := "field"
-	table := &Table{
-		Name:    tableName,
-		Columns: Columns{columnName: Column{"text"}},
-	}
-	event := map[string]interface{}{
-		columnName: "value 42",
-	}
-	eventContext := &EventContext{
-		ProcessedEvent: event,
-		Table:          table,
-	}
-
-	if err := ar.CreateTable(table); err != nil {
-		return err
-	}
-
-	defer func() {
-		if err := ar.DeleteTable(table); err != nil {
-			// Suppressing error because we need to check only write permission
-			logging.Warnf("Cannot remove table [%s] from Redshift: %v", tableName, err)
-		}
-	}()
-
-	if err := ar.Insert(eventContext); err != nil {
-		return err
-	}
-
-	return nil
 }
--- a/server/adapters/aws_redshift.go
+++ b/server/adapters/aws_redshift.go
@@ -326,6 +326,10 @@ func (ar *AwsRedshift) ValidateWritePermission() error {
 	event := map[string]interface{}{
 		columnName: "value 42",
 	}
+	eventContext := &EventContext{
+		ProcessedEvent: event,
+		Table:          table,
+	}
 
 	if err := ar.CreateTable(table); err != nil {
 		return err
@@ -338,7 +342,7 @@ func (ar *AwsRedshift) ValidateWritePermission() error {
 		}
 	}()
 
-	if err := ar.Insert(table, event); err != nil {
+	if err := ar.Insert(eventContext); err != nil {
 		return err
 	}
 
--- a/server/adapters/aws_redshift.go
+++ b/server/adapters/aws_redshift.go
@@ -331,14 +331,15 @@ func (ar *AwsRedshift) ValidateWritePermission() error {
 		return err
 	}
 
+	defer func() {
+		if err := ar.DeleteTable(table); err != nil {
+			// Suppressing error because we need to check only write permission
+			logging.Warnf("Cannot remove table [%s] from Redshift: %v", tableName, err)
+		}
+	}()
+
 	if err := ar.Insert(table, event); err != nil {
 		return err
-	}
-
-	if err := ar.DeleteTable(table); err != nil {
-		logging.Warnf("Cannot remove table %s from redshift: %v", tableName, err)
-		// Suppressing error because we need to check only write permission
-		// return err
 	}
 
 	return nil
--- a/server/adapters/aws_redshift.go
+++ b/server/adapters/aws_redshift.go
@@ -3,11 +3,14 @@ package adapters
 import (
 	"context"
 	"fmt"
-	"github.com/jitsucom/jitsu/server/logging"
-	"github.com/jitsucom/jitsu/server/typing"
-	_ "github.com/lib/pq"
+	"math/rand"
 	"strconv"
 	"strings"
+
+	"github.com/jitsucom/jitsu/server/logging"
+	"github.com/jitsucom/jitsu/server/timestamp"
+	"github.com/jitsucom/jitsu/server/typing"
+	_ "github.com/lib/pq"
 )
 
 const (
@@ -200,6 +203,16 @@ func (ar *AwsRedshift) CreateTable(tableSchema *Table) error {
 	return ar.dataSourceProxy.createTableInTransaction(wrappedTx, tableSchema)
 }
 
+// DeleteTable removes database table with name provided in Table representation
+func (ar *AwsRedshift) DeleteTable(table *Table) error {
+	wrappedTx, err := ar.OpenTx()
+	if err != nil {
+		return err
+	}
+
+	return ar.dataSourceProxy.deleteTableInTransaction(wrappedTx, table)
+}
+
 //Update one record in Redshift
 func (ar *AwsRedshift) Update(table *Table, object map[string]interface{}, whereKey string, whereValue interface{}) error {
 	columns := make([]string, len(object), len(object))
@@ -301,4 +314,32 @@ func (ar *AwsRedshift) recreateNotNullColumnInTransaction(wrappedTx *Transaction
 //Close underlying sql.DB
 func (ar *AwsRedshift) Close() error {
 	return ar.dataSourceProxy.Close()
+}
+
+func (ar *AwsRedshift) ValidateWritePermission() error {
+	tableName := fmt.Sprintf("test_%v_%v", timestamp.NowUTC(), rand.Int())
+	columnName := "field"
+	table := &Table{
+		Name:    tableName,
+		Columns: Columns{columnName: Column{"text"}},
+	}
+	event := map[string]interface{}{
+		columnName: "value 42",
+	}
+
+	if err := ar.CreateTable(table); err != nil {
+		return err
+	}
+
+	if err := ar.Insert(table, event); err != nil {
+		return err
+	}
+
+	if err := ar.DeleteTable(table); err != nil {
+		logging.Warnf("Cannot remove table %s from redshift: %v", tableName, err)
+		// Suppressing error because we need to check only write permission
+		// return err
+	}
+
+	return nil
 }
--- a/server/adapters/bigquery.go
+++ b/server/adapters/bigquery.go
@@ -222,6 +222,10 @@ func (bq *BigQuery) ValidateWritePermission() error {
 	event := map[string]interface{}{
 		columnName: "value 42",
 	}
+	eventContext := &EventContext{
+		ProcessedEvent: event,
+		Table:          table,
+	}
 
 	if err := bq.CreateTable(table); err != nil {
 		return err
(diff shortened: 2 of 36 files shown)


server/adapters/google_cloud_storage.go
```diff
@@ -32,7 +32,7 @@ type GoogleConfig struct {
 	credentials option.ClientOption
 }
 
-func (gc *GoogleConfig) Validate(streamMode bool) error {
+func (gc *GoogleConfig) Validate(strict, streamMode bool) error {
```
xtreding, line 35: Do we really need strict parameter?
`Dataset` can be empty (it has default value 'default')
`Project` is validated on BigQuery storage creation.

server/adapters/postgres.go
```diff
@@ -90,6 +94,9 @@ func (dsc *DataSourceConfig) Validate() error {
 	if dsc.Db == "" {
 		return errors.New("Datasource db is required parameter")
 	}
+	if strict && dsc.Schema == "" {
```
xtreding, line 97: Same. Schema isn't required parameter. It has default value (see Postgres/Redshift storage)

server/adapters/snowflake.go
```diff
@@ -53,7 +57,7 @@ type SnowflakeConfig struct {
 }
 
 //Validate required fields in SnowflakeConfig
-func (sc *SnowflakeConfig) Validate() error {
+func (sc *SnowflakeConfig) Validate(strict bool) error {
```
xtreding, line 60: Same. Schema and Password aren't required

server/adapters/snowflake.go
```diff
@@ -176,6 +186,31 @@ func (s *Snowflake) CreateTable(tableSchema *Table) error {
...
+// DeleteTable removes database table with name provided in Table representation
+func (s *Snowflake) DeleteTable(table *Table) error {
+	wrappedTx, err := s.OpenTx()
+	if err != nil {
+		return err
+	}
+
+	query := fmt.Sprintf(deleteSFTableTemplate, s.config.Schema, reformatValue(table.Name))
+	s.queryLogger.LogDDL(query)
+	createStmt, err := wrappedTx.tx.PrepareContext(s.ctx, query)
```
xtreding, line 198: We can use here wrappedTx.tx.ExecContext(s.ctx, query).
Without creating prepared statement

server/adapters/aws_redshift.go
```diff
@@ -302,3 +315,36 @@ func (ar *AwsRedshift) recreateNotNullColumnInTransaction(wrappedTx *Transaction
 func (ar *AwsRedshift) Close() error {
 	return ar.dataSourceProxy.Close()
 }
+
+func (ar *AwsRedshift) ValidateWritePermission() error {
```
xtreding, line 319: After moving test eventContext creation outside storage, this func can be omitted.

server/adapters/bigquery.go
```diff
@@ -197,6 +212,39 @@ func (bq *BigQuery) Close() error {
 	return bq.client.Close()
 }
 
+func (bq *BigQuery) ValidateWritePermission() error {
```
xtreding, line 215: After moving test eventContext creation outside storage, this func can be omitted.

server/adapters/clickhouse.go
```diff
@@ -612,6 +646,42 @@ func (ch *ClickHouse) reformatValue(v interface{}) interface{} {
 	return v
 }
 
+func (ch *ClickHouse) ValidateWritePermission() error {
```
xtreding, line 649: After moving test eventContext creation outside storage, this func can be omitted.

server/adapters/postgres.go
```diff
@@ -689,6 +720,39 @@ func (p *Postgres) buildQueryPayload(valuesMap map[string]interface{}) (string,
 	return strings.Join(header, ", "), strings.Join(placeholders, ", "), values
 }
 
+func (p *Postgres) ValidateWritePermission() error {
```
xtreding, line 723: After moving test eventContext creation outside storage, this func can be omitted.

server/adapters/snowflake.go
```diff
@@ -328,6 +363,39 @@ func (s *Snowflake) columnDDL(name string, column Column) string {
 	return fmt.Sprintf(`%s %s`, reformatValue(name), sqlColumnTypeDDL)
 }
 
+func (s *Snowflake) ValidateWritePermission() error {
```
xtreding, line 366: After moving test eventContext creation outside storage, this func can be omitted.

server/handlers/destinations.go
```diff
@@ -172,13 +249,73 @@ func testDestinationConnection(config *storages.DestinationConfig) error {
...
 
 		return nil
+
 	default:
-		return errors.New("unsupported destination type " + config.Type)
+		return fmt.Errorf("Unsupported destination type: '%v'", config.Type)
 	}
 }
+
+func TestProcessing(config *storages.DestinationConfig) error {
```
xtreding, line 265: It would be better to give a name like `TestBatch`

server/handlers/destinations.go
```diff
@@ -24,18 +29,20 @@ func DestinationsHandler(c *gin.Context) {
...
 	err := testDestinationConnection(destinationConfig)
 	if err != nil {
 		c.JSON(http.StatusBadRequest, middleware.ErrResponse(err.Error(), nil))
 		return
 	}
+
 	c.Status(http.StatusOK)
 }
 
 func testDestinationConnection(config *storages.DestinationConfig) error {
```
xtreding, line 42: I suppose this func should contain:
1. Creating test eventContext
2. Creating adapter or storage (depend on destination mode - stream or batch)
3. Call test funcs depend on destination mode:
    - stream: create table, insert, delete table
    - batch: TestBatchProcessing()

server/logging/factory.go
```diff
@@ -64,6 +64,10 @@ func (f *Factory) CreateIncomingLogger(tokenID string) *AsyncLogger {
 }
 
 func (f *Factory) CreateFailedLogger(destinationName string) *AsyncLogger {
+	if f == nil {
```
xtreding, line 67: It is a bad practice to have such checks and return nil. We should have dummy implementation of these structs for overriding behavior, if we don't need a common one.

server/storages/clickhouse.go
```diff
@@ -284,3 +284,7 @@ func (ch *ClickHouse) getAdapters() (*adapters.ClickHouse, *TableHelper) {
 	num := rand.Intn(len(ch.adapters))
 	return ch.adapters[num], ch.chTableHelpers[num]
 }
+
+func (ch *ClickHouse) TestBatchProcessing(testName string, events []map[string]interface{}) error {
```
xtreding, line 288: TestBatchProcessing can't return not supported error if destination supports batch mode.

Imagine the situation: User have created ClickHouse destination in batch mode and wants to test connection. Incomprehensible error will be returned, and it is a bug.

E.g. we could run stream test connection pipeline (as they are identical)

server/storages/postgres.go
```diff
@@ -260,3 +260,7 @@ func (p *Postgres) Close() (multiErr error) {
 
 	return
 }
+
+func (p *Postgres) TestBatchProcessing(testName string, events []map[string]interface{}) error {
```
xtreding, line 264: same, like ClickHouse

server/storages/types.go
```diff
@@ -30,6 +30,8 @@ type Storage interface {
 	SyncStore(overriddenDataSchema *schema.BatchHeader, objects []map[string]interface{}, timeIntervalValue string) error
 	Update(object map[string]interface{}) error
 	Fallback(events ...*events.FailedEvent)
+	TestBatchProcessing(testName string, events []map[string]interface{}) error
```
xtreding, line 33: It's a good idea to have 1 Storage func for testing connection. 
I suggest to change this func signature to:
`TestBatchProcessing(eventCtx *adapters.EventContext) error` and call this func with prepared table(with name `jitsu_test_connection_${unique_identitifer}`) that has 1 test data object.

server/storages/types.go
```diff
@@ -41,6 +43,7 @@ type Storage interface {
 //StorageProxy is a storage proxy
 type StorageProxy interface {
 	io.Closer
+	Create() (Storage, error)
```
xtreding, line 46: StorageProxy should have such func

server/handlers/destinations.go
```diff
@@ -172,13 +249,73 @@ func testDestinationConnection(config *storages.DestinationConfig) error {
...
+
+	storageProxy, _, err := factory.Create(testName, *config)
+	if err != nil {
+		logging.Errorf("[%s] Error initializing destination of type %s: %v", testName, config.Type, err)
+		return err
+	}
+
+	defer storageProxy.Close()
+
+	storage, err := storageProxy.Create()
```
xtreding, line 301: Instead of `Create()` func, storageProxy can have func like `GetStatus` or `GetError` which are return current error if a storage tried to initialize and got an error. Here we could wait initialization success or error and then do the rest of the logic.

xtreding (changes requested):

server/storages/proxy.go
```diff
@@ -87,3 +88,7 @@ func (rsp *RetryableProxy) Close() error {
 	}
 	return nil
 }
+
+func (rsp *RetryableProxy) Create() (Storage, error) {
+	return rsp.factoryMethod(rsp.config)
```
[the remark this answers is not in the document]
Alessar: It is used for `destination.TestProcessing` function where it is needed to create storage object and do tests.
But it cannot be moved into storage package and function `Create` can be make internal for package like `create`.

server/adapters/google_cloud_storage.go
Alessar: `handlers.testDestinationConnection` needs such `strict` argument. In another case, a connection will be failed.
It is needed to provide default values into adapters (for example `Project` or `Dataset`) but it is provided in the storage layer and we create duplication in this case.
Another thing that some of the values like `schema` or `database` is related to database objects which are created during preparing storage to send events. But create schemas or databases for testing connection - is a strange thing because we need to remove them after tests or in another case tests can make a lot of created schemas and databases.
So I vote that the adapter layer should be without default values, and it should be passed explicitly.
Default values can be setup in config layers or in the user layer on JS side for make connection testing.

server/adapters/snowflake.go
Alessar: Changed for using without a statement

server/adapters/aws_redshift.go
Alessar: Function was removed

server/adapters/postgres.go
Alessar: Function was removed

server/adapters/clickhouse.go
Alessar: Function was removed

server/adapters/snowflake.go
Alessar: Function was removed

server/handlers/destinations.go
Alessar: Changed to `TestBatchProcessing`

server/storages/clickhouse.go
Alessar: Refactoring made `batch` testing available for `ClickHouse` storage type.

server/storages/postgres.go
Alessar: Refactoring made `batch` testing available for `Postrgres` storage type.

server/adapters/bigquery.go
Alessar: Function was removed

server/storages/types.go
Alessar: After refactoring single `TestBatchProcessing` function exists. It does all needed operation with objects.

server/handlers/destinations.go
Alessar: Connection testing is used in a process that was invoked by the HTTP handler.
In common cases, the HTTP calls have a timeout for waiting for an answer.
So in that case, there is no time for waiting while storage will be initialized after several minutes.
So I vote that in that place storage will be created immediately without any timeouts.

server/adapters/google_cloud_storage.go
Alessar: Reverted

server/adapters/snowflake.go
Alessar: Reverted

server/adapters/postgres.go
Alessar: Reverted

server/handlers/destinations.go
Alessar: Currently function invokes `TestBatchProcessing` and `TestStreamProcessing` for most types of destinations.
That two functions contain full batch and stream logic for processing events (and logic for cleaning temporary objects).

server/logging/factory.go
Alessar: Reverted

PR #446, opened 2021-07-16
feature/38-gdpr-compliant

xtreding:
added gdpr support
make default IP lookup for s2s endpoint


server/events/js_processor.go
```diff
@@ -23,6 +26,12 @@ func (jp *JsProcessor) Preprocess(event Event, reqContext *RequestContext) {
 			logging.Warnf("Unable to set user-agent from header to event object: %v", err)
 		}
 	}
+
```
absorbb, line 29: doc to the methods is outdated

server/handlers/pixel.go
```diff
@@ -126,7 +126,13 @@ func (ph *PixelHandler) parseEvent(c *gin.Context) (events.Event, error) {
 //extractOrSetAnonymID if no anoymous id found:
```
absorbb, line 126: better rename method to extractOrSetAnonymCookieID

Issue #351, opened 2021-05-11
Library for JavaScript templates


vklimontovich:
### Problem

Currently, a lot of configuration parameters relies on [go templates](https://golang.org/pkg/text/template/). It's not very convenient: a) it's a weird language for our audience b) the syntax is way not obvious. Example:

```
{{if .metric_type }}{{ .metric_type }}_test{{else}}{{ .app }}_web_test {{end}}
```

### Solution

It would be way more easier if we allow to use JavaScript as a template language. See how example above will look in JS:

```javascript
metric_type || (app + "web_test)
```

Other examples: 

Go template (from segment config)
```
   {{if or (eq .event_type "user_identify") (eq .event_type "identify")}}
          {{"identifies"}}
        {{else}}
          {{if or (eq .event_type "page") (eq .event_type "pageview")}}
            {{"pages"}}
          {{else}}
            {{.event_type}}
          {{end}}
        {{end}}
```

Javascript
```javascript
if (event_type === 'user_identify' || event_type === 'identify') {
  return 'identifies'
} else if (event_type === 'page' || event_type === 'page_view') {
  return 'pages'
} else {
  return event_type
}
```
Javascript (alternative syntax - the result is the last evaluated expression)
```
(event_type === 'user_identify' || event_type === 'identify') ? 
  'identifies' : 
  ((event_type === 'page' || event_type === 'page_view') ? 'pages' : event_type)
```

Go syntax should be similar to go templates:

```go
type Inventory struct {
	Material string
	Count    uint
}
sweaters := Inventory{"wool", 17}
tmpl, err := jtemplate.New("test").Parse("Count > 0 ? (Material || "standard" )")
if err != nil { panic(err) }
err = tmpl.Execute(os.Stdout, Inventory{"wool", 17})
if err != nil { panic(err) }
```

Also, we need to implement `jtemplate.SmartParse("name", "expression")`. The function should return another function that executes the template. The parsing order should be as follows:
 - Try to parse `expression` as go template. If parsing is successful, check if the template is just a static string, and return `StaticTemplate` in that case
 - If parsing failed, try parse string as JavaScript
 - If JS parsing failed, return  both errors
(We need to experiment with direct order of attempts on real use-cases)

Here are preliminary go types:

```go
type TemplateExecutor (func (event events.Event) string);

func StaticTemplate(expression string) TemplateExecutor {
	return func(event events.Event) string {
		return expression;		
	}
}

[...]

PR #449, opened 2021-07-17
#351 run js functions in their own goroutine

absorbb:


server/templates/template_executor.go
```diff
@@ -42,16 +49,33 @@ func (gte *goTemplateExecutor) Format() string {
...
+
+type asyncJsTemplateExecutor struct {
+	incoming chan events.Event
+	results chan interface{}
+	transformedExpression string
+	loadingError error
 }
 
-func newJsTemplateExecutor(expression string, extraFunctions template.FuncMap) (*jsTemplateExecutor, error) {
+type pooledTemplateExecutor struct {
```
xtreding, line 73: is it used anywhere? If not, we shouldn't keep it in the codebase

server/templates/template_executor.go
```diff
@@ -63,12 +87,70 @@ func newJsTemplateExecutor(expression string, extraFunctions template.FuncMap) (
...
+
+
+func (jte *asyncJsTemplateExecutor) start(extraFunctions template.FuncMap) {
+	//loads javascript into new vm instance
+	function, err := LoadTemplateScript(jte.transformedExpression, extraFunctions)
 	if err != nil {
-		return nil, fmt.Errorf("js loading error: %v", err)
+		jte.loadingError =  fmt.Errorf("%s: %v\ntransformed function:\n%v\n",jsLoadingErrorText, err, jte.transformedExpression)
+	}
+	for {
```
xtreding, line 108: For such cases we have `safego` package that contains `safego.RunWithRestart()` func. Please take a look at the example https://github.com/jitsucom/jitsu/blob/master/server/adapters/http.go#L98

xtreding (approved):

server/templates/template_executor.go
absorbb: not used. will deleted it

server/templates/template_executor.go
absorbb: done. but for entire start() function.

PR #467, opened 2021-07-27
#456 dbt Cloud integration.

### LINKED ISSUE #456: Trigger dbt cloud job on successful source or batch destination run
### Problem

Optimize dbt cloud automation by triggering of dbt cloud Job after source successfully loads new data to destinations
or file uploader completes loading batch to destinations

### Solution

Add dbt settings to UI: toggle, accountId, JobId, ApiToken
When dbt enabled - successful completion of any source  or batch destination run must call "Trigger job to run" method of dbt API:
https://docs.getdbt.com/dbt-cloud/api#operation/triggerRun

absorbb:
UI: Added support for 'hidden' destination types
Closes #456


server/adapters/dbtcloud.go
```diff
@@ -0,0 +1,167 @@
...
+	method        = "POST"
+	urlFormat     = "https://cloud.getdbt.com/api/v2/accounts/%d/jobs/%d/run/"
+	testUrlFormat = "https://cloud.getdbt.com/api/v2/accounts/%d/jobs/%d/"
+	bodyFormat    = `{"cause": "%s"}`
+)
+
+//DbtCloudConfig is a dto for parsing DbtCloud configuration
+type DbtCloudConfig struct {
+	AccountId int    `mapstructure:"accountId" json:"accountId,omitempty" yaml:"account_id,omitempty"`
+	JobId     int    `mapstructure:"jobId" json:"jobId,omitempty" yaml:"job_id,omitempty"`
```
xtreding, line 23: I suggest using snake case in all serialization field names, like in all DTOs of such purpose.

server/adapters/dbtcloud.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+//Validate returns err if invalid
+func (dcc *DbtCloudConfig) Validate() error {
+	if dcc == nil {
+		return errors.New("DbtCloud config is required")
+	}
+	switch {
+	case dcc.AccountId == 0:
+		return errors.New("'AccountId' is required parameter")
```
xtreding, line 36: It would be better to use the serialization field name here. Otherwise, users will see this error and might be confused about field names.

server/destinations/service.go
```diff
@@ -138,6 +138,18 @@ func (s *Service) GetDestinationByID(id string) (storages.StorageProxy, bool) {
 	return unit.storage, true
 }
 
+func (s *Service) GetDestinationsByType(typ string) []storages.StorageProxy {
+	s.RLock()
+	defer s.RUnlock()
+	var res = make([]storages.StorageProxy, 0)
```
xtreding, line 144: it could be just: 
```
res := make([]storages.StorageProxy, 0)
```

server/synchronization/task_executor.go
```diff
@@ -197,6 +198,44 @@ func (te *TaskExecutor) execute(i interface{}) {
 		te.handleError(task, taskLogger, msg, true)
 		return
 	}
+	te.onSuccess(task, taskLogger)
+}
+
+func (te *TaskExecutor) onSuccess(task *meta.Task, taskLogger *TaskLogger) {
+	var dbtClouds = te.destinationService.GetDestinationsByType(storages.DbtCloudType)
```
xtreding, line 205: We should filter by project ID as well. Usually, it happened in the Configurator backend, where we just set destination IDs into source configuration.

xtreding (changes requested):

server/synchronization/task_executor.go
absorbb: Ouch. Thanks

xtreding (approved):

configurator/frontend/src/stores/destinations.ts
```diff
@@ -168,6 +171,14 @@ class DestinationsStore implements IDestinationsStore {
     return this._destinations;
   }
 
+  public get hiddenDestinations() {
```
taletski, line 174: It is a good practice to add the new public methods to the class signature to make those methods type safe. Just add `hiddenDestinations: DestinationData[]` and `allDestinations: DestinatinonData[]` to the `IDestinationsStore`

configurator/frontend/src/ui/components/ConfigurableFieldsForm/ConfigurableFieldsForm.tsx
```diff
@@ -166,6 +168,10 @@ const ConfigurableFieldsForm = ({
       const fieldsValue = form.getFieldsValue();
 
       switch (type?.typeName) {
+        case 'description':
+          return (
+              <div style={{paddingTop: 5}} dangerouslySetInnerHTML={{__html: defaultValue}}/>
```
taletski, line 173: For unification purposes, it is preferable to use tailwind classes instead of the inline styles. Consider `className="pt-2"` instead.

Also, I will attempt to get the same result without setting the html.

configurator/frontend/src/ui/pages/DestinationsPage/partials/DestinationEditor/DestinationEditor.tsx
```diff
@@ -166,7 +170,7 @@ const DestinationEditor = ({
       } catch (errors) {
         // ToDo: check errors count for fields with few validation rules
         tab.errorsCount = errors.errorFields?.length;
-        throw errors;
+        return null;
```
taletski, line 173: I think it would be better to re-throw the errors here, and if they bother you in your component just catch them there.

taletski (changes requested):
The update is neat, but there are some points we should address before merging

configurator/frontend/src/catalog/destinations/lib/amplitude.tsx
```diff
@@ -74,6 +75,6 @@ const amplitudeDestination = {
       )
     }
   ]
-} as const;
+} as Destination;
```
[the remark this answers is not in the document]
taletski: fixed

configurator/frontend/src/ui/pages/DestinationsPage/partials/DestinationEditor/DestinationEditor.tsx
absorbb: here it triggered Uncaught error in Promise in some cases. I didn't find better way to resolve that

configurator/frontend/src/ui/components/ConfigurableFieldsForm/ConfigurableFieldsForm.tsx
absorbb: Could find right solution. Left as is

configurator/frontend/src/ui/pages/DestinationsPage/partials/DestinationEditor/DestinationEditor.tsx
taletski: I see, this function is not the commonly used one, but it is used just locally in the component. Since it already gives the user feedback it is safe to leave it as is.

configurator/frontend/src/ui/components/ConfigurableFieldsForm/ConfigurableFieldsForm.tsx
taletski: fixed in aae40a1

taletski (approved):

Issue #418, opened 2021-06-25
Singer improvements (processing Singer's replication method)


xtreding:
### Problem

At present Jitsu doesn't take into account Singer's `forced-replication-method` (`INCREMENTAL` or `FULL_TABLE`). Data is deduplicated by the primary key. In case when 3rd-party platform deletes data, Jitsu will keep all data. We should improve it.

### Solution

- [ ] Singer output consumer should delete all data from table (delete * from collection.GetTableName()) if Singer's `forced-replication-method`= `FULL_TABLE`
- [ ] Clear cache endpoint should have query parameter (e.g. `delete_warehouse_data=true`). If true Jitsu should delete all data from tables.

PR #483, opened 2021-08-03
#418 cleaning for FULL_TABLE singer source replication

efgefg0001:
issue #418

* Clear cache endpoint should have query parameter (e.g. delete_warehouse_data=true). If true Jitsu should delete all data from tables. - **COMPLETED**

* Singer output consumer should delete all data from table (delete * from collection.GetTableName()) if Singer's forced-replication-method= FULL_TABLE - **COMPLETED**


server/handlers/sources.go
```diff
@@ -79,6 +86,25 @@ func (sh *SourcesHandler) ClearCacheHandler(c *gin.Context) {
 	c.JSON(http.StatusOK, middleware.OKResponse())
 }
 
+func (sh *SourcesHandler) deleteWareHouseData(driver driversbase.Driver, destinationIds []string, sourceID string, collection string, multiErr error) {
```
xtreding, line 89: Enriching input error doesn't work properly. You should explicitly return multiErr.

server/synchronization/singer_result_saver.go
```diff
@@ -87,6 +87,12 @@ func (rs *ResultSaver) Consume(representation *singer.OutputRepresentation) erro
 		rowsCount := len(stream.Objects)
 		//Sync stream
 		for _, storage := range rs.destinations {
+			if representation.NeedClean {
+				err := storage.Clean(stream.BatchHeader.TableName)
+				if err != nil {
+					logging.Warn("Storage cleaning failed, ignoring: %v", err)
```
xtreding, line 93: please add destination ID preffix like: logging.Warn("[%s] storage cleaning failed, ignoring: %v", storage.ID(), err)

server/singer/parser.go
```diff
@@ -84,7 +87,7 @@ func StreamParseOutput(stdout io.ReadCloser, consumer PortionConsumer, logger lo
 			}
 
 			outputPortion.State = state
-
+			outputPortion.NeedClean = !initialized && strings.Contains(string(lineBytes), "FULL_TABLE")
```
xtreding, line 90: It seems that not state, but [catalog](https://github.com/singer-io/getting-started/blob/master/docs/SPEC.md#catalog) contains replication method (incremental or full table)

xtreding (changes requested):

server/synchronization/singer_result_saver.go
efgefg0001: changed log message

server/handlers/sources.go
efgefg0001: Added multiErr return statement

server/singer/parser.go
efgefg0001: fixed

server/adapters/adapter.go
```diff
@@ -10,6 +10,7 @@ type SQLAdapter interface {
 	PatchTableSchema(schemaToAdd *Table) error
 	BulkInsert(table *Table, objects []map[string]interface{}) error
 	BulkUpdate(table *Table, objects []map[string]interface{}, deleteConditions *DeleteConditions) error
+	Clean(tableName string) error
```
xtreding, line 13: It would be better to have more clear name e.g. `Truncate`

server/adapters/bigquery.go
```diff
@@ -236,6 +239,14 @@ func (bq *BigQuery) DropTable(table *Table) error {
 	return nil
 }
 
+//Clean deletes all records in tableName table
+func (bq *BigQuery) Clean(tableName string) error {
+	query := fmt.Sprintf(truncateBigQueryTemplate, bq.config.Project, bq.config.Dataset, tableName)
+	bq.queryLogger.LogQuery(query)
+	_, err := bq.client.Query(query).Read(bq.ctx)
```
xtreding, line 246: here we should check returned error. If a table doesn't exist BigQuery returns a specific error.
Jitsu should omit such error and return nil.

server/adapters/clickhouse.go
```diff
@@ -450,6 +452,29 @@ func (ch *ClickHouse) BulkInsert(table *Table, objects []map[string]interface{})
...
+	ch.queryLogger.LogDDL(query)
+
+	_, err = wrappedTx.tx.ExecContext(ch.ctx, query)
+
+	if err != nil {
+		return fmt.Errorf("Error cleaning [%s] table: %v", tableName, err)
+	}
+
+	if ch.cluster != "" {
+		ch.dropDistributedTableInTransaction(wrappedTx, tableName)
```
xtreding, line 472: use cleanDistributedTableInTransaction func instead

server/adapters/mysql.go
```diff
@@ -255,11 +256,39 @@ func (m *MySQL) toDeleteQuery(conditions *DeleteConditions) (string, []interface
...
 
+//Clean deletes all records in tableName table
+func (m *MySQL) Clean(tableName string) error {
+	wrappedTx, err := m.OpenTx()
+	if err != nil {
+		return err
+	}
+
+	if err := m.cleanTableInTransaction(wrappedTx, tableName); err != nil {
+		wrappedTx.Rollback()
```
xtreding, line 267: here we don't have anything to rollback, therefore we can just use `m.datasource.ExecContext(...)` instead of using OpenTx, Commit.

server/adapters/postgres.go
```diff
@@ -258,6 +258,21 @@ func (p *Postgres) Insert(eventContext *EventContext) error {
...
 }
 
+//Clean deletes all records in tableName table
+func (p *Postgres) Clean(tableName string) error {
+	wrappedTx, err := p.OpenTx()
+	if err != nil {
+		return err
+	}
+
+	if err := p.cleanTableInTransaction(wrappedTx, tableName); err != nil {
```
xtreding, line 268: here the same as in MySQL. We don't need separate transaction creation here.

server/adapters/mysql.go
```diff
@@ -255,11 +256,39 @@ func (m *MySQL) toDeleteQuery(conditions *DeleteConditions) (string, []interface
...
 //Close underlying sql.DB
 func (m *MySQL) Close() error {
 	return m.dataSource.Close()
 }
 
+func (m *MySQL) cleanTableInTransaction(wrappedTx *Transaction, tableName string) error {
+	statement := fmt.Sprintf(mySQLCleanTableTemplate, m.config.Db, tableName)
+	m.queryLogger.LogDDL(statement)
+
+	_, err := wrappedTx.tx.ExecContext(m.ctx, statement)
```
xtreding, line 283: here the same as in BigQuery. We should check `error doesn't exist` error and omit it.

server/adapters/postgres.go
```diff
@@ -564,6 +579,19 @@ func (p *Postgres) bulkMergeInTransaction(wrappedTx *Transaction, table *Table,
 	return p.dropTableInTransaction(wrappedTx, tmpTable)
 }
 
+func (p *Postgres) cleanTableInTransaction(wrappedTx *Transaction, tableName string) error {
+	statement := fmt.Sprintf(postgresCleanTableTemplate, p.config.Schema, tableName)
+	p.queryLogger.LogDDL(statement)
+
+	_, err := wrappedTx.tx.ExecContext(p.ctx, statement)
```
xtreding, line 586: please check error about table doesn't exist

server/adapters/snowflake.go
```diff
@@ -386,6 +387,21 @@ func (s *Snowflake) DropTable(table *Table) error {
...
 }
 
+//Clean deletes all records in tableName table
+func (s *Snowflake) Clean(tableName string) error {
+	wrappedTx, err := s.OpenTx()
+	if err != nil {
+		return err
+	}
+
+	if err := s.cleanTableInTransaction(wrappedTx, tableName); err != nil {
```
xtreding, line 397: we don't need separate transaction here

xtreding (changes requested):

server/drivers/singer/singer.go
```diff
@@ -55,6 +55,8 @@ type Singer struct {
 	tableNamePrefix  string
 	streamTableNames map[string]string
 
+	StreamReplication map[string]string
```
xtreding, line 58: can be unexported (streamReplication map[string]string)

xtreding (changes requested):

server/adapters/adapter.go
efgefg0001: renamed method

server/adapters/clickhouse.go
efgefg0001: removed transaction

server/adapters/mysql.go
efgefg0001: removed transaction

server/adapters/postgres.go
efgefg0001: removed transaction

server/adapters/snowflake.go
efgefg0001: removed transaction

server/schema/flattener.go
```diff
@@ -91,6 +91,10 @@ func Reformat(key string) string {
 	return result.String()
 }
 
+func TableName(prefix, streamName string) string {
```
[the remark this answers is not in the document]
efgefg0001: removed TableName function

server/drivers/singer/singer.go
efgefg0001: to lower case

PR #494, opened 2021-08-09
Fix asyncJsTemplateExecutor goroutines leak after destination reload

absorbb:


xtreding (approved):

xtreding:
Please, also, consider using `closed atomic.Bool` instead of `closed chan struct{}` like in `adapters/http.go`. 
I suppose using `atomic.Bool` is more laconic due to less code.

absorbb:
> Please, also, consider using `closed atomic.Bool` instead of `closed chan struct{}` like in `adapters/http.go`.
> I suppose using `atomic.Bool` is more laconic due to less code.

The thing is that closing channel automatically breaks for select loop in goroutine. that won't happen with Bool

PR #509, opened 2021-08-29
Implementation of time freezing

### LINKED ISSUE #503: Tests should be compatible with Apple Silicon (M1 chip)
## Description
Jitsu uses [bouk/monkey](https://github.com/bouk/monkey) lib to freeze time in tests, but the library doesn't work properly with Apple Silicon.

## Steps to reproduce

Run `./server/make test_backend` on Apple M1 macbook.

## Expected behavior

All tests are passed

## Actual behavior

Tests are failed

## Proposed implementation

- [ ] All time.Now() calls should be replaced by a custom interface that can be mocked in tests.
- [ ] Remove bouk/monkey library from tests

Alessar:
Fixes #503 

* Own implementation of getting the current time
* `time.Now()` -> `timestamp.Now()`
* Own implementation of freezing and unfreezing time

### DIFF
--- a/server/timestamp/timestamp.go
+++ b/server/timestamp/timestamp.go
@@ -1,26 +1,30 @@
 package timestamp
 
-import "time"
+import (
+	"time"
+
+	"go.uber.org/atomic"
+)
 
 var (
 	// The value of freezed time that is used in all tests
 	freezedTime = time.Date(2020, 06, 16, 23, 0, 0, 0, time.UTC)
 
 	// Indicator shows that time was freezed or was not freezed
-	timeFreezed = false
+	timeFreezed = atomic.NewBool(false)
 )
 
 func Now() time.Time {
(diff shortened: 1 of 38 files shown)


server/timestamp/timestamp.go
```diff
@@ -0,0 +1,26 @@
+package timestamp
+
+import "time"
+
+var (
+	// The value of freezed time that is used in all tests
+	freezedTime = time.Date(2020, 06, 16, 23, 0, 0, 0, time.UTC)
+
+	// Indicator shows that time was freezed or was not freezed
+	timeFreezed = false
```
xtreding, line 10: please use atomic boolean here

xtreding (approved):

server/timestamp/timestamp.go
Alessar: Changed

Issue #558, opened 2021-09-27
SNOWFLAKE DESTINATION , add gcp k8s workload identity support


viggin543:
### Problem

When running jitsu on GKE (google k8s engine) , it is possible to assign k8s service accounts google cloud engine roles.
for example access to google cloud storage
This mechanism eliminates the need to pass service account JSON keys around. 
Once a pod assumes a service account it has access to all resources allowed by the google service account.
This works out of the box for all google SDK / CLI tools.
Currently, when not providing a service account key when creating a snowflake destination the google sdk client fails ( since its probably given an empty string as a service account json )
![image](https://user-images.githubusercontent.com/6231756/134955520-54cd3345-1aec-40f9-855b-64fd1d0f1da9.png)

### Solution

when creating a google cloud storage client, Just don't provide any key, google cloud sdk will attempt to use the default service account installed on the pod, and google workload identity will do its magic.
should be a super simple thing to do.

vklimontovich:
Hi @viggin543! The PR on this would be very welcome :) We don't have an operational k8s cluster to test this feature

viggin543:
> Hi @viggin543! The PR on this would be very welcome :) We don't have an operational k8s cluster to test this feature

https://github.com/jitsucom/jitsu/pull/561

absorbb:
@viggin543 i've merged PR to master.
it will be part of today's hotfix release

absorbb:
Thank you for the PR
Released here [v1.35.5](https://github.com/jitsucom/jitsu/releases/tag/v1.35.5)

viggin543:
This is how a snowflake gcs configuration should look like.

![image](https://user-images.githubusercontent.com/6231756/135105410-f1325456-6b9b-45e1-8e63-a58b3e54c25f.png)
![image](https://user-images.githubusercontent.com/6231756/135105521-908db2a4-ae52-4878-8df8-0817376f4cc2.png)

![image](https://user-images.githubusercontent.com/6231756/135105656-25dae686-88db-4b7b-86a8-3d2f1f34cbb1.png)
when hovering over the SA KEY input ( the image above )
A tooltip appears . And leads to a documentation page.

absorbb:
@viggin543  Thank you for the PR ad docs update
Released in Jitsu 1.35.7

PR #563, opened 2021-09-29
workload_identity support for big query

viggin543:
big query adaptor contractor used nil credentials, which causes a panic, in a scenario of a workload_identity configuration, now it creates a big query client without any creds which cause the SDK to use the default service account installed on the machine


viggin543:
Follow up PR to issue [558](https://github.com/jitsucom/jitsu/issues/558)

PR #564, opened 2021-09-29
Fix for heroku deployment

absorbb:


commandodev:
@absorbb - just to let you know, I pulled this branch and deployed it and it still hasn't fixed the issues I was having with heroku...

PR #653, opened 2021-11-08
array type

efgefg0001:
store arrays as:
* `jsonb` for postgres (aws_redshift)
* `json` for mysql
* `Array(String)` for Clickhouse
* `Array<String>` for bigquery
*  `array` for snowflake


absorbb:
Not need for now. Closing

PR #711, opened 2021-12-09
added separate redis for user recognition

xtreding:


docker-compose.yml
```diff
@@ -8,6 +8,8 @@ services:
     image: jitsucom/jitsu
     environment:
       - REDIS_URL=redis://redis:6379
+      - USER_RECOGNITION_ENABLED=true
```
absorbb, line 11: should it be enabled?

vklimontovich: Yeah, in docker compose it should. We consider compose as a "demo" of Jitsu, so no harm can be done.

But we need to add comments — users will make infra using compose as a reference

PR #741, opened 2022-01-07
appconfig/version: fix default selection

ezradiniz:
I noticed that jitsu/server was using more CPU than usual. 
After investigating a bit, I saw that "VersionReminder" was running a for-select with a "default" case.


xtreding:
thank you @ezradiniz

PR #751, opened 2022-01-13
747 made users recognition by any identification node

xtreding:
[747 issue]: made all identification events which have at least one identification node value a sign to update anonymous events in DWH. 
fixed default recognition identification nodes - now they are: /user/email and /user/id.


absorbb:
> * Default recognition identification nodes are: `/user/email` and `/user/id`. The latter one replaces `/user/internal_id` which has been introduced by mistake

we need to leave /user/internal_id there too. otherwise it will break for all users that currently uses internal_id

PR #769, opened 2022-01-25
Reworked distributed locks implementation #762

### LINKED ISSUE #762: Reworking Locks implementation
### Problem

At present Jitsu uses [redsync](https://github.com/go-redsync/redsync) for distributed locking when updating tables/synchronizing sources/getting and saving entities from UI. There are some problems with locks releasing. Since `redsync` implementation is just locking the identifier with timeout and the library doesn't handle server critical errors such as a killing the server or restarting it. As the result, obtained locks live until the timeout. Usually, it causes such errors as:

`System error: Unable to lock <>: redsync: failed to acquire lock` or 
`System error: [...] Error getting lock source [...] collection [...] task [...]: redsync: failed to acquire lock`

We should improve it.

### Solution

- [x] Locks interfaces should look something like this:

```
type LockFactory interface {
	CreateLock(name string) Lock
}

type Lock interface {
	Lock(timeout) error
	Unlock() bool
	TryLock()
}
```

- [x] Lock instance should lock only for a short period of time. When a dedicated goroutine will be run (1 goroutine per 1 lock) which renews the lock every = `lock_time/2`.

xtreding:
This PR resolves #762: the lock error that occurs after killing Jitsu Server. In this case, the server hasn't unlocked all resources properly and after the restart can't obtain locks for a lock time (3 hours). This fix included:
- Set distributed (Redis) lock time to 8 seconds
- Added background heartbeat goroutine for every lock for extending lock every lock time / 2
- Reworked locks interfaces

Since Jitsu docker now is built based on Debian image - unused build scripts and docker files have been deleted.


vklimontovich:
I have a few concerns about the Lock interface:

```go
type Lock interface {
	Lock(timeout time.Duration) error
	Unlock() bool
	TryLock() error
}
```

`TryLock()` signature looks rather unconventional. It should return `bool`, not error. The pattern of usage is

```go
	if lock.TryLock() {
		defer lock.Unlock()
		//do stuff assuming lock is aquired
	} else {
		//alternative actions
	}
```

Also, is there any reason why `Unlock()` should return `bool`? It's not used anywhere, and I can't see any use-cases where return value could be used?

Few other things:
 * Do we really need implementation for etcd? Shouldn't we remove it at all?
 * `TryLock` should have an (optional) timeout parameter as well. It's ok to not implement that so far if we don't have a use-case. However, due to nature of Go-lang (lack of operator overloading) it would make sense to introduce the param now and implement it later

vklimontovich:
Actually, the interface should look like

```go
//Lock all operations with lock
type Lock interface {
	Unlock()
	// TryLock Attempts to acquire lock within given amount of time. If lock is not free by
	// that time, returns false. Otherwise, returns true
	TryLock(timeout time.Duration) bool
}
Locking without timeout doesn't make any practical sense
```

PR #777, opened 2022-01-27
Fix error in redis LRANGE syntax

MAH69IK:
It looks like Jitsu hasn't worked with Redis all this time. At least from the bdac1ba commit from 18.06.2021.

Redis LRANGE need two arguments, but Jitsu provide only one - https://redis.io/commands/lrange


xtreding:
@MAH69IK thank you!

PR #816, opened 2022-02-16
UI Refactor: Moves common methods of stores to the EntitiesStore superclass

taletski:
The goal is to deduplicate the implementation of common objects manipulation methods such as `patch` or `delete` in objects stores. 

As a part of this task we also aim to further reduce the use of `/v1/configuration/*` APIs by removing the methods `get` and `save` from the `BackendStorageService`.


taletski:
This has been merged together with #856

Issue #823, opened 2022-02-18
Events cache: keep entities errors separately


absorbb:
### Problem

Errors often get replaced from events cache with new events - so it is hard for clients to notice them and to react.

### Solution

- ✅ Introduce separate collections for errors only: by token and by destination.
- ✅ For token store parsing errors (see #822) with bad events payload (trim payload to some limit e.g. 10kb)
- ✅ Use rate limiter to collect errors (same as in #766)
- ✅In UI show cached errors in Statistics page of each destinations
- ✅For token introduce  Statistics page with token level metrics plot (like in #822) and errors block

taletski:
Took over this task to implement the following:

- [ ] Add the "Latest Logs" button to the destination editor, it should redirect to the tasks logs page with filters on to show logs for the destination;
- [ ] "Live Events" button in Statistics Page should redirect user to the events by the API Key that was chosen in the UI;
- [ ] Add a hint about errors happened within the last month to the tasks logs page;
- [ ] API Keys should provide the same UI with statistics charts that Destinations and Sources already do;

PR #839, opened 2022-02-24
Feature/766 cache optimization

### LINKED ISSUE #766: Events Cache performance problem
### Problem

Current events cache implementation is quite heavy and prone to performance or stability problem.
Even if we keep only last 100(capacityPerDestination) events per destination in cache we try to write each event to redis even it is 10000 per second. 
Updating event status is heavy and relies on lua script.

### Possible Solutions

1. Inmemory solution:
Consider simply keeping last `capacityPerDestination` events per destination in memory and using broadcast messages to collect them from all servers. No need to update events. We can use separate collection for statuses and do zipping on backend or frontend.
Only problem that it wont't persist restrarting - but events can be dumped at shutdown

2. Redis solution:
When events cache queue grows big we must not handle every event from it.
drain it to intermediate buffers with capacity=capacityPerDestination. So no more than `capacityPerDestination` events will be written to redis.
Write entire buffer to redis at once.
No need to update record in redis. We can use separate collection for statuses and do zipping on backend or frontend.
That also will allow to get rid of special logic of maintaining capacityPerDestination in redis in favor of simple list collection and LTRIM operation

xtreding:
This PR resolves #766 and #823:

- added rate limiting to writing events to the cache
- added separate errors collection
- changed Events cache API


server/caching/events_cache.go
```diff
@@ -2,56 +2,82 @@ package caching
...
-	poolSize               int
-	trimIntervalMs         time.Duration
-	lastDestinations       sync.Map
-	done                   chan struct{}
+	storage             meta.Storage
+	rawEventsChannel    chan *rawEvent
+	statusEventsChannel chan *statusEvent
+
+	capacityPerTokenOrDestination int
+	poolSize                      int
```
absorbb, line 22: looks like we no longer use pool size

xtreding: we use it in creating goroutines

Issue #812, opened 2022-02-14
Support of SSO via BoxyHQ layer


vklimontovich:
We need to support SSO login via [BoxyHQ](https://boxyhq.com/) OAuth → SAML proxy. The implementation should lay ground for future SSO provider integration, however at the moment we should be focused on getting BoxyHQ integration working.

## Requirements

As a result of this issue, users should be able to config BoxyHQ for self-hosted Jitsu Configurator instances (via env var, or config). If the configuration is provided, the button `Continue with SSO` should appear under login button.

So far we don't need to create a new user account if auth is successful, but the account doesn't exist

## Tech details.

The feature should be configurable via `SSO_CONFIG` variable which contains JSON:

```
{
    "provider": "boxyhq",
    "url": "..."
    "tenant": "..."
    "product": "..."
    "client_secret"
}
```

Also, it should be possible to configure SSO within configurator.yaml keeping the same structure.

## Links

[1] - BoxyHQ documentation: https://boxyhq.com/docs/jackson/saml-flow#3-oauth-20-flow

Flyover-ArtSk:
@vklimontovich we already have different providers for authorization.
Example for cloud (configurator.yaml):
```
auth:
  firebase:
    project_id: 
    credentials_file:
  admin_domain: 
  admin_users:
```
I can create new boxyhq auth provider similar to the firebase provider and use it there.

PR #844, opened 2022-03-01
#812 - documentation and fixes for sso auth feature

Flyover-ArtSk:
added sso auth feature documentation
fixed bug with redirect after authorization

### DIFF
--- a/configurator/backend/authorization/service.go
+++ b/configurator/backend/authorization/service.go
@@ -46,6 +46,7 @@ type SSOConfig struct {
 	Product               string        `json:"product" validate:"required"`
 	Host                  string        `json:"host" validate:"required"`
 	AccessTokenTTLSeconds time.Duration `json:"access_token_ttl_seconds" validate:"required"`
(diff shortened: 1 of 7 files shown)


absorbb (approved):

configurator/backend/authorization/service.go
```diff
@@ -113,6 +114,7 @@ func CreateSSOProvider(vp *viper.Viper) SSOProvider {
 			Product:               vpSSO.GetString("product"),
 			Host:                  vpSSO.GetString("host"),
 			AccessTokenTTLSeconds: vpSSO.GetDuration("access_token_ttl_seconds"),
+			ConfiguratorRootPath:  vpSSO.GetString("configurator_root_path"),
```
jfk9w, line 117: It feels like this should be a more general configuration parameter `ui.base_url`, like in server already. This would allow the configuration parameter to be repurposed (and it feels like we may want to use it elsewhere).

PR #890, opened 2022-04-12
Node ipc

jfk9w:


jfk9w:
Contains Node IPC executor for JavaScript.

PR #895, opened 2022-04-18
save active project to local storage

Flyover-ArtSk:


vklimontovich:
Let's follow up this PR with following change. UI should have access to following interface: 

```
export type UserStorage<T = any> = {
  setItem(name: string, value: T, opts?: {ttlSeconds?: number}): Promise<void>;
  getItem(name: string): Promise<T>;
  removeItem(name: string): Promise<void>;
}
```

The implementation should call jitsu backend, and backend should call redis. 

When it's implemented, we need to get rid of all `localStorage` usage. `localStorage` is good for caching data, but never as a user settings storage

PR #920, opened 2022-04-28
Singer config persistence

absorbb:
Persist mutated singer configs between task runs, test and catalog discover calls.
Fast approach was used w/o big refactoring.

Side effect:
Once source is created no change in main config will have any effect - workaround is to delete source and create new


hash-data:
what is issue that is being solved by persisting config of singer sources?

absorbb:
@hash-data some singer based source updates oath access token each run - so we need to save new values. otherwise these sources will fail authorization.

PR #928, opened 2022-05-16
Js debug log

jfk9w:
Pass script execution logs to `evaluate` response for UI code debugger.
Refactor script executor for `script.Listener`.


vklimontovich:
TODO:
- [ ] JavaScript test for sources: first test should load sources from list of npm packages names; second test should execute a simple sources that does not require any external dependency. This source (`testing-source`) should be added to [jitsu-sdk](https://github.com/jitsucom/jitsu-sdk) and published to npm too

PR #937, opened 2022-05-26
Fix incorrect path for importing sdk-js module

simple10:
The NPM module does not currently allow `import "@jitsu/sdk-js"` due to the incorrect path. This fixes it.

### DIFF
--- a/javascript-sdk/package.json
+++ b/javascript-sdk/package.json
@@ -3,7 +3,7 @@
   "version": "2.5.5",
   "description": "Jitsu JavaScript SDK (more at http://jitsu.com/docs/js-sdk)",
   "main": "dist/npm/jitsu.cjs.js",
-  "module": "dist/npm/dist/jitsu.es.js",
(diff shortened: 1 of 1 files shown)


vklimontovich:
Hi @simple10! Thank you for reporting this. We moved SDK to a [dedicated repo](https://github.com/jitsucom/jitsu-js/tree/master/packages/javascript-sdk). @Flyover-ArtSk could you please make sure that this issue is fixed there, and delete `javascript-sdk` folder here to avoid confusion?

Flyover-ArtSk:
Hi @simple10! Thank you for your RP. 
I published new version of package (2.5.6) with your fix.

vklimontovich:
Just to be clear: `2.5.5` was affected by this bug. And @Flyover-ArtSk released `2.5.6` that fixes that.

But we have new version, `3.x.x`. It's not affected, and I highly recommend to upgrade to it. It's 100% API compatible. The last one is 3.1.1

PR #940, opened 2022-05-30
Gcs destination

### LINKED ISSUE #884: Google Cloud Storage destination
There is separate destination for Amazon S3. 
But there is no destination for Google Cloud Storage - and our client needs one.
It's easy to do because we have the most of the logic implemented already for BigQuery and Snowlake.

jfk9w:
Adds GCS destination (resolves #884)


absorbb:
Please add it to UI
in /Users/ildarnurislamov/Projects/jitsu/configurator/frontend/catalog/destinations/lib
and to index.ts there too

Issue #876, opened 2022-03-29
Add more metrics


absorbb:
### Problem

We have to little metrics to be sure that application is healthy and there are no regressions after upgrading to a new version

### Solution

For every significant event that take place during lifetime add prometheus metrics: 

- app start / shutdown
- unit initialization (destination, source, api_key, etc) success | error
- new unit created (destination, source, api_key, etc) success | error
- unit test connection (destination, source) success | error
- task_executor.go: source(collection) sync run success | error | cancel | stale?
- uploader.go: api token file processed (break by destination & table) success | error 
- backend auth status success | login_failed | error
- open API endpoint statuses http codes? maybe already collected by gin
- mail send status success | error ?
- slack notification send status success | error
- replay status success | error
- ...

PR #954, opened 2022-07-23
Adding application metrics

Alessar:
Fixes issue #876
Adding some metrics to eventnative application


server/metrics/destinations.go
```diff
@@ -0,0 +1,37 @@
...
+	successDestinations *prometheus.Counter
+	errorDestination    *prometheus.Counter
+)
+
+func initDestinations() {
+	successDestinations = NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "eventnative",
+			Subsystem: "destinations",
+			Name:      "success",
```
absorbb, line 17: please clarify metrics names and corresponding methods.
init_success, init_error

That applies for sources to

server/authorization/service.go
```diff
@@ -182,12 +184,15 @@ func (s *Service) GetTokenID(tokenFilter string) string {
 func (s *Service) updateTokens(payload []byte) {
 	tokens, err := parseFromBytes(payload)
 	if err != nil {
+		metrics.ErrorTokenUpdating()
```
absorbb, line 187: Better rename both to ErrorAPIKeyUpdating SuccessAPIKeyUpdating.
That is how we used to call this entity

server/metrics/destinations.go
absorbb: Many things may happens with destinations and sources. it is not clear success of what.
that is why i propose change naming to: init_success, init_error

absorbb:
- ✅ app start / shutdown / ~~running_time~~ - its ok to bw used instead of start / shutdown
- ✅ unit initialization (~~destination~~, ~~source~~, ~~api_key~~, etc) ~~success | error~~
- ✅  ~~new unit created (destination, source, api_key, etc) success | error~~ not needed
- unit test connection (destination, source) success | error
- task_executor.go: source(collection) sync run success | error | cancel | stale?
- uploader.go: file processed (break by destination & table ) success | error
- configurator backend auth status success | login_failed | error
- open API endpoint statuses http codes - enable gin metrics for configurator (copy code from server)
- ✅ mail send status ~~success | error~~
- ✅ slack notification send status ~~success | error~~
- ✅ replay status ~~success | error~~

server/metrics/destinations.go
Alessar: Renamed

server/authorization/service.go
Alessar: Renamed

PR #1005, opened 2023-03-23
Return Auth0 sub claim instead of random UUID

pahanix:
Try to get sub claim first and fallback to random UUID but not vice versa


absorbb:
Hi @pahanix !
The code is actually correct. `MapNVLKeys` is a variadic function - it tries to get first present value from keys provided.
But since variadic argument must be the last one in the function signature we had to put 'defaultValue' as a second argument.
So `uuid.New()` is a default value that is used when there are no values for 'sub' and 'email' fields

pahanix:
@absorbb you are correct! I'm sorry! I came to a wrong conclusion.

PR #1014, opened 2023-05-17
extend sso profile with access token

alexeybokov:


absorbb:
Hi! @alexeybokov 
Thank you for the PR
Can you please share motivation behind these changes?

alexeybokov:
Hi @absorbb, in our case we want to use token for some integrations that we call from frontend side. probably it's not good idea to pass it here explicitly

PR #1084, opened 2024-02-06
fix: handle error responses from bulker in log endpoint

echozio:
Without this the API returns a 200 containing the error from the Bulker (including the token used for authentication if it's rejected by the Bulker), and an authentication failure between the console and Bulker results in a somewhat cryptic error from https://github.com/jitsucom/jitsu/blob/fb73e5921cfdaf94e487e9cbf7978279b614360d/webapps/console/components/DataView/EventsBrowser.tsx#L974 (evs.map is not a function).


absorbb:
thank you!

PR #1085, opened 2024-02-06
build: copy build output to the correct paths

echozio:
Using `console.Dockerfile` as it was resulted in a blank page and lots of 404s. Updated paths are the same as in `all.Dockerfile`:
https://github.com/jitsucom/jitsu/blob/fb73e5921cfdaf94e487e9cbf7978279b614360d/all.Dockerfile#L38-L39


absorbb:
@echozio Thank you!

PR #1092, opened 2024-03-11
fix: update segment tracking integration documentation

hendrikheil:
The segment plugin for `@analytics` has two distinct APIs. One for the browser and one for the server.

I've updated the docs to include examples for both and also updated the import style to ESM, which I think is more common now.


absorbb:
@hendrikheil 
This piece of doc is no longer in UI. 
Do you think that it is worth to bring it back?

hendrikheil:
I'm personally a fan of keeping installation examples in proper documentation anyway. I think it might be preferable to close this PR then 👍

PR #1101, opened 2024-04-20
Improve GTM integration

vklimontovich:
This PR improves GTM plugin in `@jitsu/js`. It also improves the data pushed to GTM in Console.

Plus, it fixes a few issues with user authorization for firebase

### DIFF
--- a/libs/jitsu-js/src/destination-plugins/gtm.ts
+++ b/libs/jitsu-js/src/destination-plugins/gtm.ts
@@ -15,51 +15,84 @@ function omit(obj: any, ...keys: string[]) {
 export const gtmPlugin: InternalPlugin<GtmDestinationCredentials> = {
   id: "gtm",
   async handle(config, payload: AnalyticsClientEvent) {
+    const debug = !!config.debug;
     if (!applyFilters(payload, config)) {
       return;
     }
     await initGtmIfNeeded(config, payload);
 
     const dataLayer = window[config.dataLayerName || "dataLayer"];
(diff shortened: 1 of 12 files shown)


libs/jitsu-js/src/destination-plugins/gtm.ts
```diff
@@ -7,51 +7,92 @@ export type GtmDestinationCredentials = {
   dataLayerName?: string;
 } & CommonDestinationCredentials;
 
+function omit(obj: any, ...keys: string[]) {
+  const set = new Set(keys);
```
absorbb: on small number of elements Set is actually slower than arrays. So simple keys.includes will do

webapps/console/prisma/schema.prisma
```diff
@@ -60,6 +60,9 @@ model UserProfile {
   userPreferences         UserPreferences[]
   workspaceUserProperties WorkspaceUserProperties[]
   password				  UserPassword?
+
+  @@index(externalId)
+  @@unique([loginProvider, externalId])
```
absorbb, line 65: Prisma requires `--accept-data-loss` flag here.
(currently table doesn't contain duplicates)

>   There might be data loss when applying the changes:
  • A unique constraint covering the columns `[loginProvider,externalId]` on the table `UserProfile` will be added. If there are existing duplicate values, this will fail.
Error: Use the --accept-data-loss flag to ignore the data loss warnings like prisma db push --accept-data-loss
 ELIFECYCLE  Command failed with exit code 1.

PR #1110, opened 2024-06-20
feat(destination/fb): Optionally send phone field

denisroldan:
Add support for sending the hashed phone from the user traits following [Meta spec](https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/customer-information-parameters#ph).


vklimontovich:
Thanks, @denisroldan! @absorbb will merge it by the beginning of the next week

PR #1122, opened 2024-08-28
posthog-destination: support $screen_name for screen events

jspizziri:
Currently, the `$screen_name` property isn't being passed properly to posthog on `screen` events. You can see that being done in [posthog's native integration here](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-react-native/src/posthog-rn.ts#L155)


absorbb:
Thank you!

PR #1124, opened 2024-08-30
facebook-conversions: map page & screen events to "ViewContent" events

jspizziri:
Currently `screen` and `page` events are showing up in Meta as custom events, rather than their standard `ViewContent` event which is designed for pages.

### DIFF
--- a/libs/core-functions/src/functions/facebook-conversions.ts
+++ b/libs/core-functions/src/functions/facebook-conversions.ts
@@ -37,6 +37,19 @@ function toPrettyString(responseJson: any) {
   return typeof responseJson === "string" ? responseJson : JSON.stringify(responseJson, null, 2);
 }
 
+// https://developers.facebook.com/docs/meta-pixel/reference#standard-events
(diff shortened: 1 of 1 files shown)


absorbb:
Thank you!

jspizziri:
@absorbb , has this been deployed to cloud?

absorbb:
@jspizziri not yet. It needs more testing

absorbb:
@jspizziri have to rollback it because it will affect all users already using facebook destination

jspizziri:
@absorbb interesting. I suppose that makes sense, but it is strange that people would have to manually map this in order for it to show up correctly in FB. Is there a way to add a timestamp in the code like:

```
if (destination.created_at <= new Date('...')) {
  // use old mapping
} else {
  // use new mapping
}
```

Or a feature flag or something? At the very least, it's probably wise to document this somewhere.

absorbb:
@jspizziri 
we may consider that. Also Segment uses "PageView" event for page event by default. 
So not sure what is the best default here

PR #1136, opened 2024-10-16
 Allow overriding properties url and path for page events 2

absorbb:
analytics-plugin: respect overridden 'url': make corresponding 'context' properties consistent.


libs/jitsu-js/src/analytics-plugin.ts
```diff
@@ -410,19 +410,32 @@ function adjustPayload(
...
-  const properties = payload.properties || {};
+  const referrer = runtime.referrer();
 
-  if (payload.type === "page" && url) {
+  if (properties.url) {
     properties.url = url.replace(hashRegex, "");
-    properties.path = fixPath(urlPath(url));
+  }
+  const ctxPage = {
+    path: url ? fixPath(urlPath(url)) : properties.path,
```
huoxito, line 423: wonder if this should prioritize the path provided through `properties` as well:

```js
path: properties.path || fixPath(urlPath(url))
```

huoxito (approved):
thank you! 🙌🏼

libs/jitsu-js/src/analytics-plugin.ts
```diff
@@ -410,19 +410,32 @@ function adjustPayload(
   s2s: boolean
 ): AnalyticsClientEvent {
   const runtime: RuntimeFacade = config.runtime || (isInBrowser() ? windowRuntime(config) : emptyRuntime(config));
-  const url = runtime.pageUrl();
+  const properties = payload.properties || {};
+  const url = properties.url || runtime.pageUrl();
```
huoxito, line 414: 👋🏼 just realized something else related to the context option argument in `analytics.track`

for events such as for example track "Product Viewed" we don't necessarily need a `properties.url` or `properties.path` but we do want to make sure `context.page.url` and `context.page.path` for those events are accurate.

For those "track" scenarios within Shopify iframe sandbox pixel setup we're passing the url and path as part of the `analytics.track` [third argument](https://getanalytics.io/api/#analyticstrack) called `options`.

I think jitsu could follow the same segment / analytics api behaviour, which allows us to override `context` options as well, by modifying this line to be something like:

```js
const url = properties.url || properties.context?.url || runtime.pageUrl();
```

otherwise we're only able to override the url and path under track `context` if we also override `url` at the `properties` level

absorbb: > otherwise we're only able to override the url and path under track context if we also override url at the properties level

I don't think that providing `context` in `properties` is a good way to go here. Because it is not clear whether we need to remove this object from `properties` or leave it. In both ways that is not transparent.

But support for overriding `context` via `options` third argument looks like a good idea.
Also that is the approach suggested by Segment: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/faq/#can-i-overwrite-the-context-fields.

We consider supporting for overriding context values as suggested in the link above.
At the same time it will be possible to override properties of page event

absorbb: >I don't think that providing context in properties is a good way to go here. Because it is not clear whether we need to remove this object from properties or leave it. In both ways that is not transparent.

Correction. We already support overriding `context` via `properties` and we remove that provided context from properties afterwards. So we will keep it this way

absorbb: @huoxito i will restore you original PR then.
This one has a high chance of breaking compatibility.

PR #1156, opened 2024-12-12
feat (@jitsu/js): Add `defaultPayloadContext` to include default context in all requests

dmeremyanin:
This PR introduces a new property `defaultPayloadContext` to the configuration. It allows default context data (e.g., browser timezone, some extra user identifier) to be automatically included in every request's payload.

### Example Usage

```js
window.jitsuConfig = {
  defaultPayloadContext: {
    browserTz: Intl.DateTimeFormat().resolvedOptions().timeZone,
    userIdentity: getUserIdentity(),
  }
}
```

The values of `browserTz` and `userIdentity` will now be included in the payload of every request, ensuring that the common context data is consistently available in all API calls, without needing to manually include them with each request.

### DIFF
--- a/libs/jitsu-js/src/analytics-plugin.ts
+++ b/libs/jitsu-js/src/analytics-plugin.ts
@@ -39,6 +39,7 @@ const defaultConfig: Required<JitsuOptions> = {
   s2s: undefined,
   idEndpoint: undefined,
   errorPolicy: "log",
+  defaultPayloadContext: {},
   privacy: {
     dontSend: false,
     disableUserIds: false,
@@ -445,7 +446,10 @@ function adjustPayload(
     properties.path = fixPath(urlPath(targetUrl));
   }
 
-  const customContext = payload.properties?.context || payload.options?.context || {};
+  const customContext = deepMerge(
+    config.defaultPayloadContext,
+    payload.properties?.context || payload.options?.context || {}
+  );
   delete payload.properties?.context;
   const referrer = runtime.referrer();
   const context: AnalyticsClientEvent["context"] = {
--- a/libs/jitsu-js/__tests__/node/nodejs.test.ts
+++ b/libs/jitsu-js/__tests__/node/nodejs.test.ts
@@ -214,6 +214,33 @@ describe("Test Jitsu NodeJS client", () => {
     expect((p.body.anonymousId ?? "").length).toBeGreaterThan(0);
   });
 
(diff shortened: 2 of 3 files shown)


libs/jitsu-js/src/analytics-plugin.ts
```diff
@@ -445,6 +446,7 @@ function adjustPayload(
   delete payload.properties?.context;
   const referrer = runtime.referrer();
   const context: AnalyticsClientEvent["context"] = {
+    ...config.defaultPayloadContext,
```
absorbb: it is not possible to introduce nested property like `context.page.awesomePageProp` this way.

Better to use `deepMerge` approach one the line 484 bellow

dmeremyanin: Yeah, agreed. I'll look into it.

dmeremyanin: Done. The code now supports nested objects in the `defaultPayloadContext`.

libs/jitsu-js/src/analytics-plugin.ts
```diff
@@ -483,7 +485,7 @@ function adjustPayload(
     messageId: randomId(properties.path || (parsedUrl && parsedUrl.pathname)),
     writeKey: maskWriteKey(config.writeKey),
     groupId: storage.getItem("__group_id"),
-    context: deepMerge(context, customContext),
+    context: deepMerge(defaultPayloadContext, deepMerge(context, customContext)),
```
absorbb: I guess that approach will be more clear:
Line 449:
```typescript
const customContext = deepMerge(config.defaultPayloadContext, payload.properties?.context || payload.options?.context || {});
```
and this line untouched

dmeremyanin: Thanks for the suggestion! The code is now a bit cleaner with that approach.

absorbb (approved):

dmeremyanin:
@absorbb thanks for the prompt review of all the PRs. Also, just wondering - when can we expect a new release with the latest changes?

absorbb:
Usually we make a release to fix some critical issue or when we collected significant amount of minor changes. 
But these changes are already in `beta` images. So you can switch to `beta` until we issue another release.

dmeremyanin:
Got it, thanks for the clarification. We will switch to the beta images in the meantime.

PR #1160, opened 2024-12-18
chore(@console|@profile): remove database credentials from logs

pedroyremolo:
## Why

Currently, in the console app and profiles service database credentials were being exposed in logs when a new database client connection was established. This can lead to a secret leak which is undesirable.

## What

This change removes the connection URL from logs since having this information on the connection step does not add any value and in case of failure, the error already shows its cause.

### DIFF
--- a/services/profiles/src/lib/db.ts
+++ b/services/profiles/src/lib/db.ts
@@ -197,7 +197,7 @@ export function createPg(): Pool {
     log
       .atInfo()
       .log(
-        `Connecting new client ${connectionUrl}. Pool stat: idle=${pool.idleCount}, waiting=${pool.waitingCount}, total=${pool.totalCount}` +
+        `Connecting new database client. Pool stat: idle=${pool.idleCount}, waiting=${pool.waitingCount}, total=${pool.totalCount}` +
           (schema ? `. Default schema: ${schema}` : "")
       );
     //this is commented on purpose, it won't work for pgbouncer in transaction mode https://www.pgbouncer.org/features.html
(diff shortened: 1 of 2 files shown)


vklimontovich:
Let's mask credentials, instead of removing it. E.g. print

`postgresql://username:******@host/newjitsu?sslmode=no-verify&schema=newjitsu`

instead 

`postgresql://username:SECRET_PASSWORD@host/newjitsu?sslmode=no-verify&schema=newjitsu`

We might already have a function for it somewhere. @absorbb did the masking in events logs

pedroyremolo:
> Let's mask credentials, instead of removing it. E.g. print
> 
> `postgresql://username:******@host/newjitsu?sslmode=no-verify&schema=newjitsu)`
> 
> instead

okey dokey, gonna adjust here =)

pedroyremolo:
> We might already have a function for it somewhere. @absorbb did the masking in events logs

During this research on jitsu repo I've found some options, I guess this one is suitable @vklimontovich :
https://github.com/jitsucom/jitsu/blob/newjitsu/webapps/ee-api/lib/store.ts#L85

May I replicate this function on these apps?

absorbb:
@pedroyremolo 
It is better to move this function into `libs/juava` subproject. it is common lib for all of them

pedroyremolo:
> @pedroyremolo It is better to move this function into `libs/juava` subproject. it is common lib for all of them

perfect. Gonna do it here 🙏🏾

rodfsoares:
hey @vklimontovich and @absorbb 👋 

I've opened another version of this PR - @pedroyremolo is on PTO, and I don't have permission to commit to his personal branch 😅 

I've applied your suggestions in #1164 - can we continue the discussion there, please?

PR #1164, opened 2025-01-02
chore(@console|@profile): remove database credentials from logs

rodfsoares:
(rework of #1160)

## Why
Currently, in the console app and profiles service database credentials were being exposed in logs when a new database client connection was established. This can lead to a secret leak which is undesirable.
 
## What
This change removes the connection URL from logs since having this information on the connection step does not add any value and in case of failure, the error already shows its cause.

### DIFF
--- a/libs/juava/src/security.ts
+++ b/libs/juava/src/security.ts
@@ -76,3 +76,13 @@ export function decrypt(secret: string, iv: string, value: string): string {
   const decrypted = Buffer.concat([decipher.update(Buffer.from(value, "hex")), decipher.final()]);
   return decrypted.toString();
 }
+
+export function hideSensitiveInfo(url: string) {
+  try {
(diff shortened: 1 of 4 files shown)


libs/juava/src/security.ts
```diff
@@ -76,3 +76,13 @@ export function decrypt(secret: string, iv: string, value: string): string {
   const decrypted = Buffer.concat([decipher.update(Buffer.from(value, "hex")), decipher.final()]);
   return decrypted.toString();
 }
+
+export function hideSensitiveInfo(url: string) {
```
rodfsoares, line 80: moved from `webapps/ee-api/lib/store.ts` as requested:
https://github.com/jitsucom/jitsu/pull/1160#issuecomment-2564294427

absorbb (approved):

PR #1170, opened 2025-01-17
chore(@console): remove risky console.logs

rodfsoares:
Console's logs are exposing basic auth credentials when an admin logs in. Here's an example (values manually redacted by me):
Screenshot 2025-01-17 at 10 59 48

I believe the cause is the `console.log()` in `nextauth.config.ts`. I assume this is a leftover from a debug attempt, and since it doesn't seem to offer much value in Production, I think it's best to remove it outright.

I also spotted another potentially risky instance in the SignInOrUp component, so removing that as well.

### DIFF
--- a/webapps/console/components/SignInOrUp/SignIn.tsx
+++ b/webapps/console/components/SignInOrUp/SignIn.tsx
@@ -117,7 +117,6 @@ export const SocialLogin: React.FC<{ onSocialLogin: (type: string) => Promise<vo
               await onSocialLogin("github.com");
               redirectIfNeeded(router);
             } catch (error: any) {
-              console.log(JSON.stringify(error, null, 2));
               setError(handleFirebaseError(error));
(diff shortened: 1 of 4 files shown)


webapps/console/components/SignInOrUp/SignIn.tsx
```diff
@@ -117,7 +117,6 @@ export const SocialLogin: React.FC<{ onSocialLogin: (type: string) => Promise<vo
               await onSocialLogin("github.com");
               redirectIfNeeded(router);
             } catch (error: any) {
-              console.log(JSON.stringify(error, null, 2));
               setError(handleFirebaseError(error));
```
rodfsoares, line 120: I think `handleFirebaseError()` already logs the error, so best to remove an apparently stray `console.log()`

absorbb (approved):

PR #1185, opened 2025-03-09
rotor: Enhance Redis support

dmeremyanin:
## Overview

This PR introduces key enhancements to Redis support:

1. Redis Sentinel support - automatic failover and HA for Redis instances.
2. Redis can now be used as the exclusive persistent storage solution, replacing MongoDB.
3. Metrics tracking for Redis storage.

No breaking changes introduced - it supports smooth migration from Redis to MongoDB for users who still wish to use MongoDB.

## Motivation

Our workload has shown that Redis outperforms MongoDB significantly in terms of resource efficiency. Specifically, MongoDB was consuming nearly 1 CPU core per replica and 750MB of memory at just ~250 requests per second, which we believe is excessive for such a simple use case as K/V storage.

Although Jitsu's documentation states that Redis is deprecated in favor of MongoDB, I believe this decision should be reconsidered. Based on our internal performance tests, Redis offers a more efficient solution than MongoDB for our workload. Redis' resource usage is far more optimal, making it a more suitable choice in scenarios with high-performance demands.

Supporting both MongoDB and Redis would provide greater flexibility, allowing users to choose the best storage solution for their needs. For users who don't require features like identity stitching and prefer Redis for its speed and resource efficiency, this approach would also result in more cost-effective Jitsu deployments.

### DIFF
--- a/services/rotor/src/lib/redis.ts
+++ b/services/rotor/src/lib/redis.ts
@@ -18,29 +18,66 @@ function hideSensitiveInfoFromURL(url: string) {
   return parsed.toString();
 }
 
-export function createRedis(): Redis {
-  const redisUrl = requireDefined(process.env.REDIS_URL, "env REDIS_URL is not defined");
-  log.atDebug().log(`Building redis client for ${hideSensitiveInfoFromURL(redisUrl)}`);
-  let tls: any = undefined;
+function resolveRedisConnectionOptions(
+  redisUrl: string,
+  redisSentinelAddress: string | undefined
+): Record<string, any> {
+  let sentinels: any;
+
+  if (redisSentinelAddress) {
+    sentinels = redisSentinelAddress.split(",").map(sentinel => {
+      const [host, port] = sentinel.split(":");
+      return {
+        host,
+        port: port ? parseInt(port, 10) : undefined,
+      };
+    });
+  }
+
+  let tls: any;
   if (redisUrl.startsWith("rediss://")) {
     tls = {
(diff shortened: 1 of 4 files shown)


services/rotor/src/lib/redis.ts
```diff
@@ -18,29 +18,63 @@ function hideSensitiveInfoFromURL(url: string) {
...
+}
+
+/**
+ * Example `REDIS_URL` values for a standalone Redis instance:
+ * - Standalone Redis: `redis://username:password@localhost:6379/1`
+ * - Standalone Redis with SSL: `rediss://username:password@localhost:6379/2`
+ *
+ * Example `REDIS_URL` and `REDIS_SENTINEL_ADDRESS` values for Redis Sentinel:
+ * - Redis URL: `redis://username:password@/3?name=mymaster`
+ * - Redis Sentinel Address: `sentinel1:26379,sentinel2:26379,sentinel3:26379`
```
dmeremyanin, line 60: A separate `REDIS_SENTINEL_ADDRESS` env variable is required because `ioredis` doesn't support configuring Redis Sentinel directly through the connection URL. Redis Sentinel requires a list of sentinel addresses to be passed separately.

absorbb (approved):

absorbb (approved):

absorbb:
@dmeremyanin 
Thank you. Especially for the performance data

absorbb:
also `beta` image is ready

PR #1186, opened 2025-03-14
De-duplicate getSeed generation in high throughput scenarios

ikraamg:
Adds a counter that increments with each call to getSeed()
This ensures that even if multiple calls happen at the exact same timestamp,
they will receive different seed values due to the incrementing counter.

Reasons for this solution:
- Isomorphic compatibility (No external dependencies)
- Simple but effective
- The implementation is lightweight and doesn't introduce significant overhead.
- Using `% Number.MAX_SAFE_INTEGER` ensures the counter stays within JavaScript's safe integer bounds

### DIFF
--- a/libs/jitsu-js/src/analytics-plugin.ts
+++ b/libs/jitsu-js/src/analytics-plugin.ts
@@ -942,9 +942,14 @@ export const jitsuAnalyticsPlugin = (jitsuOptions: JitsuOptions = {}, storage: P
   };
 };
 
+let seedCounter = 0;
 function getSeed() {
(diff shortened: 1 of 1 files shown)


absorbb:
Thank you.
Just curious. We have two `Math.random()` calls in `randomId` method
Do you have some cases when it is not enough?

ikraamg:
I am not exactly sure why because theoretically it seems we should not have an issue.

During some load tests (millions of events in a short timeframe) from a server to these endpoints `:url/api/s/:event_type`, we had duplicates when we relied on the Jitsu. Solved by passing in an externally generated UUID per request. 
It is probably also also caused by having hashString the same for all events we sent.

absorbb (approved):

PR #1187, opened 2025-03-24
Avail PropertyUI types and hide sensitive fields in the destination configuration

ikraamg:
- Make the PropertyUI type available in the destination configuration by moving it over (not sure about how to best do this).
- Hide sensitive fields in the destination configuration UI.

### DIFF
--- a/libs/core-functions/src/lib/meta-types.ts
+++ b/libs/core-functions/src/lib/meta-types.ts
@@ -1,0 +1,48 @@
+/**
+ * UI for property
+ */
+export type PropertyUI = {
+  /**
+   * Optional human-friendly name of the field
+   */
+  displayName?: string;
+  /**
+   * If string field should be treated as textarea (multiline input)
+   */
+  textarea?: boolean;
+  /**
+   * If string field should be treated as password
+   */
+  password?: boolean;
+  /**
+   * If the field should not be displayed. That field must have a default value
+   */
+  hidden?: boolean | ((obj: any) => boolean);
+  /**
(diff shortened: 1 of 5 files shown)


libs/core-functions/src/lib/meta-types.ts
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Name of custom editor component. See getEditorComponent() function from `[workspaceId]/destinations.txt`
+   */
+  editor?:
+    | "ArrayTextarea"
+    | "StringArrayEditor"
+    | "CodeEditor"
+    | "SnippedEditor"
+    | "MultiSelectWithCustomOptions"
+    | string;
```
ikraamg, line 43: I have added these types for ease of reference. They look to be the only ones currently supported. I have not removed the acceptance of the `string` type since there is a runtime check for the editor type and I wanted to tread lightly with this change.

webapps/console/lib/schema/destinations.tsx
```diff
@@ -28,7 +30,6 @@ import intercomIcon from "./icons/intercom";
 import webhookIcon from "./icons/webhook";
 import { branding } from "../branding";
 import * as meta from "@jitsu/core-functions/src/meta";
-import { HubspotCredentials } from "@jitsu/core-functions/src/meta";
```
ikraamg, line 31: This was already available via `meta.HubspotCredentials`

danhemerlein (approved):

PR #1190, opened 2025-04-01
Identity Stitching improvements

absorbb:
rotor: user-recognition: by default initiate stitching only by 'identify' even with non-empty userId
rotor: user-recognition: support IDENTIFYING_TRAITS functions env var to initiate stitching by present traits
rotor: functions chain optimizations
profile-builder: add identity stitching to transformation pipeline


vklimontovich (approved):
LGTM

PR #1192, opened 2025-04-08
rotor: performance optimizations

absorbb:
rotor: don't do excessive event copy if functions pipeline consists of just 1 function
rotor: added support for 'fastFunctions' option that disables copying of events between functions chain steps
rotor: bulker-destination: switched to undici
rotor: small tweaks


vklimontovich (approved):
LGTM

PR #1193, opened 2025-04-08
Fix/event browser html

danhemerlein:
This updates some classNames to use a string (no need for {} if no dynamic logic)
If we're not using dynamic logic or variables, just pass the string directly

This also moves the EventsBrowser to a directory in DataView. I did this so I could move some types into their own file. I would like do another push of cleaning up this file (breaking it into smaller components)

Thank you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ node_modules
 /cli/build-scripts/dist/
 /libs/jsondiffpatch/coverage/
 /libs/jsondiffpatch/dist/
+/devenv-*
--- a/webapps/console/components/DataView/DataView.tsx
+++ b/webapps/console/components/DataView/DataView.tsx
@@ -96,6 +96,7 @@ export function DataView() {
       destroyInactiveTabPane={true}
       type="line"
       items={items}
+      className="w-full"
     />
   );
 }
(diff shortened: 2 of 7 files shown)


vklimontovich:
Although this PR totally make sense, we generally hesitant of making cosmetic changes and moving files around without strong reason to do so in scope of adding new features.

Speaking of `classNames`, that would make sense if we could add a prettier/linter rule to enforce that.

Looks like `react/jsx-curly-brace-presence` does exactly this, I'll try to add via separate commit

danhemerlein:
> Although this PR totally make sense, we generally hesitant of making cosmetic changes and moving files around without strong reason to do so in scope of adding new features.
> 
> Speaking of `classNames`, that would make sense if we could add a prettier/linter rule to enforce that.
> 
> Looks like `react/jsx-curly-brace-presence` does exactly this, I'll try to add via separate commit

All good thank you very much for the review @vklimontovich please feel free to close this out if it doesn't make sense. Happy to help out with new feature development - especially if it's console app related. thanks!

vklimontovich:
Here's the change: https://github.com/jitsucom/jitsu/commit/ccbc8bfb54d2750ef11633ae26a808ab696841ac

We will swap it from `warn` to `error` once all PRs are merged to avoid merge conflicts

PR #1196, opened 2025-04-28
feature: soft delete workspace

danhemerlein:
This adds UI and an API endpoint Delete operation.
The Delete operation soft deletes workspaces only if they have no connections. 

This also adds the `confirmOpWithInput` which requires triggers a confirmation modal requiring the user input a specific string before enabling the 'OK' button.

Please see this screen recording demonstrating the new functionality:
[https://share.zight.com/qGu8lrPD](https://share.zight.com/qGu8lrPD)

A big thing to keep in mind here is that database records are not actually deleted - they're soft deleted in the database
    

---

## Summary by mrge
Added the ability to delete a workspace from the settings page, with a confirmation flow and backend support. Workspaces can only be deleted if they have no active connections.

- **New Features**
  - "Delete workspace" button in the workspace settings, with a multi-step confirmation and required input.
  - API endpoint to soft delete a workspace if it has no connections.


webapps/console/components/JitsuButton/JitsuButton.tsx
```diff
@@ -10,6 +10,7 @@ import { useWorkspace } from "../../lib/context";
 export type JitsuButtonProps = ButtonProps & {
   //set to true if href is relative workspace link
   ws?: boolean;
+  className?: string;
```
danhemerlein, line 13: this allows for additional css classes to be added to buttons

vklimontovich:
Overall, it looks good, thank you very much for the contribution!

But we can make it even better by addressing a few issues:

* Delete workspace button is too bright. We don't need to capture our user attention with this section.

A good example would be Vercel's UI. Button is still red, but other

CleanShot 2025-04-28 at 15 49 10@2x


CleanShot 2025-04-28 at 15 39 59@2x

* When user clicks on Cancel for this dialog, the workspace still got deleted

CleanShot 2025-04-28 at 15 56 29@2x

danhemerlein:
> Overall, it looks good, thank you very much for the contribution!
> 
> But we can make it even better by addressing a few issues:
> 
> * Delete workspace button is too bright. We don't need to capture our user attention with this section.
> 
> A good example would be Vercel's UI. Button is still red, but other
> 
> CleanShot 2025-04-28 at 15 49 10@2x
> 
> CleanShot 2025-04-28 at 15 39 59@2x
> 
> * When user clicks on Cancel for this dialog, the workspace still got deleted
> 
> CleanShot 2025-04-28 at 15 56 29@2x

Thanks for this great feedback - will address this shortly!

danhemerlein:
> Overall, it looks good, thank you very much for the contribution!
> 
> But we can make it even better by addressing a few issues:
> 
> * Delete workspace button is too bright. We don't need to capture our user attention with this section.
> 
> A good example would be Vercel's UI. Button is still red, but other
> 
> CleanShot 2025-04-28 at 15 49 10@2x
> 
> CleanShot 2025-04-28 at 15 39 59@2x
> 
> * When user clicks on Cancel for this dialog, the workspace still got deleted
> 
> CleanShot 2025-04-28 at 15 56 29@2x

addressed this feedback!

absorbb:
added check that user actually has access to workspace being deleted.
removed requirement that all connection must be deleted - that requirement makes it hard to restore workspace to the original state
removed requirement that user must have >1 workspaces - sometimes we get asked to delete only workspace that user has

absorbb (approved):

PR #1209, opened 2025-07-11
Feat/newjitsu/roles

absorbb:
## Summary by cubic
Sensitive fields like passwords and API keys are now masked in API responses and UI for destinations and services, and are not exposed to users. When updating configs, masked values are ignored so secrets are not overwritten unless changed.

- **New Features**
  - Added logic to detect and mask secret fields for destinations and services.
  - Implemented helpers to remove masked values from updates, preventing accidental secret erasure.
  - Added tests for secret masking and removal logic.


vklimontovich (approved):
LGTM

PR #1230, opened 2025-11-26
feat(ingest): add configurable TRUSTED_IP_HEADERS for proxy support

VolanDeVovan:
When running Jitsu behind Cloudflare, the service couldn't get the real client IP because Cloudflare uses CF-Connecting-IP header instead of the standard X-Real-Ip or X-Forwarded-For.

  This PR adds a **TRUSTED_IP_HEADERS** environment variable that lets you configure which headers to check for client IP, in priority order.

  Why a configurable list instead of just adding Cloudflare support?

  In real deployments, you might have mixed scenarios:
  - Frontend behind Cloudflare sending browser events
  - Backend services (e.g., Next.js server) sending server-side events through different proxies
  - Custom proxy setups with their own headers

  A simple "enable Cloudflare" flag wouldn't work here. The configurable header list gives flexibility for any proxy setup.

### DIFF
--- a/bulker/ingest/config.go
+++ b/bulker/ingest/config.go
@@ -63,6 +63,13 @@ type Config struct {
 	ShutdownTimeoutSec int `mapstructure:"SHUTDOWN_TIMEOUT_SEC" default:"10"`
 	//Extra delay may be needed. E.g. for metric scrapper to scrape final metrics. So http server will stay active for an extra period.
 	ShutdownExtraDelay int `mapstructure:"SHUTDOWN_EXTRA_DELAY_SEC" default:"5"`
+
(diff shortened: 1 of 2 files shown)


vklimontovich:
From the design stand-point, LGTM. @absorbb would you please take a look?

absorbb:
@VolanDeVovan 
Just curious
Cloudflare by default should add X-Forwarded-For header along with CF-Connecting-IP
https://developers.cloudflare.com/fundamentals/reference/http-headers/#x-forwarded-for

Could you please make sure that there is no Managed Transform rule that removes X-Forwarded-For header?

PR #1249, opened 2026-03-13
fix(function-server) run fs on deno runtime

absorbb:
fix(operator,function-server) remove entities config and code after compiling functions
fix(function-server) free deployment: use deno web workers to isolate workspace functions from each other


vklimontovich (approved):
LGTM

Issue #1243, opened 2026-02-24
jitsucom/rotor does not start without the ClickHouse envs


Ogoyukin:
## Summary

`jitsucom/rotor` fails to start in self-hosted deployments if ClickHouse environment variables are not defined, even when ClickHouse is not intended to be used.

When `jitsu-clickhouse-secret` is removed (or `CLICKHOUSE_HOST` / `CLICKHOUSE_URL` are not set), the rotor pod goes into `CrashLoopBackOff` with:

```
[32m2026-02-24 06:42:12.539Z INFO  [kafka-rotor]: Subscribing to kafka topics:  ["destination-messages"][0m
[31m2026-02-24 06:42:12.574Z ERROR [rotor]: Failed to start rotor processing [0m
[31m Error: env CLICKHOUSE_HOST is not defined[0m
[31m     at toError (/app/main.js:385414:12)[0m
[31m     at requireDefined (/app/main.js:385440:11)[0m
[31m     at clickhouseHost (/app/main.js:388887:90)[0m
[31m     at createMetrics (/app/main.js:388570:10)[0m
[31m     at _callee8$ (/app/main.js:406384:25)[0m
[31m     at /app/main.js:406198:1001[0m
[31m     at Generator.<anonymous> (/app/main.js:406204:1670)[0m
[31m     at Generator.next (/app/main.js:406205:377)[0m
[31m     at rotor_asyncGeneratorStep (/app/main.js:406215:76)[0m
[31m     at _next (/app/main.js:406216:168)[0m
[32m2026-02-24 06:42:12.575Z INFO  [kafka-rotor]: Closing kafka-rotor [0m
[32m2026-02-24 06:42:12.576Z INFO  [kafka-rotor]: Kafka-rotor closed gracefully. 💜 [0m
[32m2026-02-24 06:42:12.576Z INFO  [rotor]: Process exited with code 1 [0m
```

From the code analysis, the reason appears to be unconditional ClickHouse initialization inside `createMetrics()` during service startup.

Although event logging already supports a fallback (`DummyEventsStore`) when ClickHouse is not configured, metrics initialization still requires ClickHouse configuration and fails early. This makes ClickHouse effectively mandatory for rotor startup.

For self-hosted environments, this creates an unexpected hard dependency on ClickHouse.

---

## System configuration and versions

- Jitsu Next (newjitsu)
- Deployment type: self-hosted (Kubernetes)
- Image: `jitsucom/rotor:<tag>`
[...]

PR #1251, opened 2026-03-19
Return dummy metrics if ClickHouse host or URL is not set in Rotor

jberlyn:
## Problem

When running without ClickHouse (e.g. local dev without the `clickhouse` service), rotor crashes on startup with `CLICKHOUSE_HOST is not defined`. This happens because `createMetrics()` unconditionally calls `getClickhouseConfig()`, which throws when neither `CLICKHOUSE_HOST` nor `CLICKHOUSE_URL` is set.

This issue was raised by someone here:
https://github.com/jitsucom/jitsu/issues/1243

## Fix

`eventsLogger` already has a fallback for this case — it uses `DummyEventsStore` when ClickHouse env vars are absent (`src/index.ts:77-81`). This PR applies the same pattern to metrics by adding an early return in `createMetrics()` that yields the already-exported `DummyMetrics` when ClickHouse isn't configured.

## AI Usage Disclosure

I used Claude Code to analyze the rotor codebase and identify the root cause and the existing pattern used for `eventsLogger`. The fix and this description are my own.


redcat-berlynj:
@vklimontovich What is the correct way to get this reviewed?

absorbb (approved):

PR #1254, opened 2026-04-09
Use FETCH_TIMEOUT_MS instead of a hardcoded 2000ms

### LINKED ISSUE #1253: FETCH_TIMEOUT_MS env var ignored in FunctionsHandlerMulti — hardcoded 2000ms fallback
## Summary

Setting `FETCH_TIMEOUT_MS` (e.g. in Helm values) has no effect when events are processed through the `FunctionsHandlerMulti` HTTP handler. This handler falls back to a hardcoded `2000` ms instead of reading the env var, so users always see 2000ms timeouts regardless of their configuration.

All other timeout-sensitive paths in Rotor (`rotor.ts`, `builder.ts`, `profiles-functions.ts`, `functions-server.ts`) correctly use `parseNumber(env.FETCH_TIMEOUT_MS, 2000)`. `FunctionsHandlerMulti` in `services/rotor/src/http/functions.ts` (line 40) was missed.

## Steps to reproduce

1. Set `FETCH_TIMEOUT_MS=10000` in Rotor env (e.g. Helm values)
2. Send events that trigger destination functions
3. Observe 2000ms timeouts — the env var value is not respected

## Fix

Import `parseNumber` + `getServerEnv` in `functions.ts` and replace the hardcoded `2000` with `parseNumber(getServerEnv().FETCH_TIMEOUT_MS, 2000)`.

jberlyn:
- Replaces the hardcoded `2000`ms fallback timeout in the functions HTTP handler with a configurable value read from the `FETCH_TIMEOUT_MS` environment variable.
- The default remains `2000ms` if the env var is not set, preserving existing behavior.

In `FunctionsHandlerMulti`, when the `x-request-timeout-ms` header is not present on a request, the fetch timeout fell back to a hardcoded `2000`. This value is now read from `getServerEnv().FETCH_TIMEOUT_MS` at startup (using `parseNumber` with a default of `2000`), making it configurable without a code change.

Fixes #1253

### DIFF
--- a/services/rotor/src/http/functions.ts
+++ b/services/rotor/src/http/functions.ts
@@ -1,4 +1,4 @@
-import { getLog, requireDefined } from "juava";
+import { getLog, parseNumber, requireDefined } from "juava";
 import { IngestMessage } from "@jitsu/protocols/async-request";
 import { MessageHandlerContext, rotorMessageHandler } from "../lib/message-handler";
 import { CONNECTION_IDS_HEADER } from "../lib/rotor";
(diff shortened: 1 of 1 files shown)


absorbb (approved):

absorbb:
We plan to deprecate this part by moving user function processing to a dedicated component.
However, I don’t see any issue with merging this for now.

PR #1273, opened 2026-04-20
feat: allow configuring clusterMode

KadekM:
Destination currently always emits `ENGINE = Replicated*MergeTree('/clickhouse/tables/{shard}/<db>/<table>', '{replica}') `when cluster is set. 

That works for classic multi-node ClickHouse, but it's rejected by the server when the target database itself uses `ENGINE=Replicated(...)`  and in that mode ClickHouse manages the ZooKeeper path and replica macros automatically and refuses explicit arguments.


KadekM:
@absorbb this would help our setup, or do you have some alternative ideas?

bulker/bulkerlib/implementations/sql/clickhouse.go
```diff
@@ -141,18 +142,26 @@ const (
 	ClickHouseProtocolHTTPS  ClickHouseProtocol = "https"
 )
 
+type ClickHouseClusterMode string
```
absorbb: It would be more universal to change this parameter to `ClickHouseDatabaseEngine`
with possible values:
default - Atomic or Shared (for Clickhouse Cloud)
replicated - Replicated

bulker/bulkerlib/implementations/sql/clickhouse.go
absorbb, line 452: If `replicated` engine is selected in config - we need to respect this setting when creating a new DB if one doesn't exist

absorbb:
@KadekM
That option makes significant changes on how bulker works with ClickHouse
Full test suit must be run for this configuration.
Please check [bulker/bulkerlib/implementations/sql/bulker_test.go:31 ](https://github.com/jitsucom/jitsu/blob/831985c5a9d7de74920f0b038fb383aac1c79aad/bulker/bulkerlib/implementations/sql/bulker_test.go#L31)
`ClickHouseBulkerTypeId + "_cluster"` config and how it is configured bellow.
Similar docker compose test container must be setup for an additional test configuration with Replicated db engine

KadekM:
Thanks, I'll look into it!

PR #1298, opened 2026-05-11
feat(services): use slack-notify reusable workflow, accept slack_thread_ts

vklimontovich:
## Summary

- Adds optional `slack_thread_ts` input (workflow_call + workflow_dispatch). Cross-repo callers (e.g. `jitsu-cloud-infra`'s deploy.yaml) that already opened a Slack thread can pass the parent ts so this build's release notification posts as a reply.
- Moves the release Slack notification out of `finalize_release` into a new `notify_slack` job that calls `jitsucom/github-workflows/.github/workflows/slack-notify.yml@main` — a reusable workflow can only be invoked at the job level.
- Exposes `deploy_run_url` as a `finalize_release` job output so the new notify job can reference it.

## Dependencies

Requires **`jitsucom/github-workflows#<PR>`** to be merged first (the reusable workflow needs the new inputs).

Pinned to `@main` for now — bump to the next tag in a follow-up after github-workflows cuts one.

## Test plan

- [ ] Manual `services.yaml` dispatch on a non-canary channel: verify the slack notification still posts (in `#dev` for now, not threaded).
- [ ] Cross-repo: run `jitsu-cloud-infra` deploy with the matching PR; verify the release notification posts into the deploy thread.

### DIFF
--- a/.github/workflows/services.yaml
+++ b/.github/workflows/services.yaml
@@ -34,6 +34,11 @@ on:
         required: false
         type: boolean
         default: false
+      slack_thread_ts:
+        description: "If set, the release Slack notification is posted into this thread (parent's ts). Used by cross-repo callers that already opened a parent message."
+        required: false
+        type: string
+        default: ""
   workflow_dispatch:
     inputs:
       channel:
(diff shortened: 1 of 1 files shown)


vklimontovich:
Addressed review feedback in b851b0d:
- `notify_slack` now fires on success **and** failure (gated by `!cancelled() && !inputs.dont_publish && needs.version.outputs.channel != 'canary'` instead of requiring `finalize_release.result == 'success'`). Header, color, and the workflow-run link branch on the final result.
- Reusable workflow pinned from `@main` → `@1.19.20260511`.

PR #1331, opened 2026-05-26
fix(js-sdk): identify user on init when userId is present

### LINKED ISSUE #1250: userId not POSTed despite existing in config
## Summary

The userId property is not sent to Jitsu (hosted), but is returned by jitsu.getConfiguration(). 
It is also not returned by jitsu.getState().
My assumption is that the HTML tag is successfully updating the Configuration object, but not State.
If jitsu.identify() is called, State is correctly updated.

## System configuration and versions

This is the hosted version of Jitsu.

## Artifacts (logs, etc)

Setup is as follows:

* jitsu base script installed through GTM (google tag manager)
* userId set as a prop on the <script> element

jagannalla:
### Summary
This PR fixes Jitsu Issue #1250 where the browser JS SDK parsed the `userId` from HTML script data-attributes (`data-user-id` or `window.jitsuConfig.userId`) into the configuration object but never automatically registered the user state or sent it with early events on startup.

### Root Cause
In `libs/jitsu-js/src/browser.ts`, script attributes successfully parsed `data-user-id` into `getConfiguration().userId`. However, in `libs/jitsu-js/src/index.ts`'s `createUnderlyingAnalyticsInstance()`, the SDK never checked for `opts.userId` during startup, leading to:
1. `jitsu.page()` and early events being fired without the user ID.
2. `jitsu.user().id` (via `jitsu.getState()`) returning `undefined` until a manual `identify()` was called.

### Changes Made
1. **Type Definitions:** Added `userId?: string` option to `JitsuOptions` in `types/protocols/analytics.d.ts` for first-class type safety.
2. **Auto-Identification Logic:**
   - Pre-set `__user_id` inside the storage wrapper right before the underlying `analytics` package instantiates, allowing `client.user().id` to return the configured ID **synchronously** on startup.
   - Automatically trigger `a.identify()` during SDK startup to dispatch the identify payload to the server.
3. **Unit Tests:** Added a robust test suite `test auto-identification on init` inside `libs/jitsu-js/__tests__/node/nodejs.test.ts` verifying all scenarios (configuration storage, user state population, auto-triggering identify dispatch, and automatic attachment to subsequent track events).

### Verification
Ran the unit test suite inside an isolated Docker container:
- All **12 unit tests passed with 100% success** (duration: 13.29s).

Closes #1250


jagannalla:
Thanks for the review. I've pushed a new commit addressing both points:

1. Unhandled Rejection: Added an autoIdentify option flag to the startup identify call so that initialization-time failures are caught and logged as warnings instead of throwing process-crashing rejections under errorPolicy: "rethrow".
2. Privacy/Reset Leak: Updated the user() fallback method to return a copy of the state instead of mutating in-place, and added checks to ensure the initial userId is not returned if reset() is called or if privacy flags are enabled.

Added corresponding unit tests for these cases and verified that the entire test suite passes. Let me know if everything looks good!

absorbb:
Fixed in `@jitsu/js:1.11.0-beta.76` and `jitsucom/console:2.14.0-beta.123`

jagannalla:
Hi @absorbb, thank you for merging the PR!

PR #1357, opened 2026-06-14
fix(juava): use CSPRNG for randomId() to prevent token prediction (#1…

### LINKED ISSUE #1356: Predictable security tokens from `Math.random()` (CWE-338)
## Scope

This report covers one defect: the token generator `randomId()` draws its characters from `Math.random()`, a non-cryptographic source. It covers the affected call sites inside `webapps/console`, the exploit path, a proposed fix, and proposed regression tests. It does not cover two other findings from the same audit - an open redirect in `webapps/console/pages/index.tsx` and a timing-unsafe token comparison in `libs/juava/src/security.ts` - which warrant separate reports.

## Summary

| Field | Value |
|---|---|
| Identifier | CWE-338 (Use of Cryptographically Weak PRNG) |
| Component | `libs/juava` (`@jitsu/juava`) |
| Defective function | `randomChar()` in `libs/juava/src/id.ts` |
| Public entry point | `randomId()` |
| Impact | An attacker who observes a few generated identifiers can predict subsequent identifiers, including bearer credentials |
| Proposed remediation | Route `randomChar()` through a cryptographically secure source (see [Proposed fix](#proposed-fix)) |

## The defect

`randomId()` is the project's single token generator. It assembles a string one character at a time by calling `randomChar()`. The current `randomChar()` selects each character with `Math.random()`:

```js
function randomChar(noDigits?: boolean) {
  const chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
  while (true) {
    const index = Math.floor(Math.random() * chars.length);
    if (!noDigits || index > 9) {
      return chars[index];
    }
  }
}
```

`Math.random()` is a pseudo-random number generator, not a cryptographically secure one. V8 implements it with the xorshift128+ algorithm. xorshift128+ holds 128 bits of internal state and emits a deterministic sequence from that state. An attacker who collects a short run of consecutive outputs can solve for the internal state, then reproduce every prior and future output. Published solvers recover the state from as few as five consecutive `Math.random()` doubles.

The generator's output length does not compensate for the weak source. `randomId(32)` produces 32 characters drawn from a 62-character alphabet, which would carry about 190 bits of entropy from a secure source. The realized entropy is bounded by the 128-bit PRNG state, and that state is recoverable, so the practical entropy against an attacker who has seen earlier outputs is near zero.

## Why it is exploitable

`randomId()` mints bearer credentials. A bearer credential is a secret string that grants access to whoever presents it, with no second factor. The following call sites generate such credentials:

- `webapps/console/pages/api/user/keys.ts:46-47` generates an API key secret and id with `randomId(32)`.
- `webapps/console/pages/api/user/cli-key.ts:17-18` generates a CLI key secret with `randomId(32)` and an id with `randomId(22)`.
- `webapps/console/pages/api/workspace/[workspaceIdOrSlug]/users/index.ts:118` generates a workspace invitation token with `randomId(12)`.

The exploit path runs as follows. The attacker first collects several identifiers that the same process emitted. Public identifiers suffice, because `randomId()` draws all identifiers from one shared `Math.random()` stream. Examples of public identifiers include object ids returned by the API. The attacker then recovers the xorshift128+ state from those samples. With the state, the attacker predicts the next outputs of the stream. When a victim then creates an API key or receives a workspace invitation, the attacker computes the secret without ever seeing it. The attacker presents the predicted secret and authenticates as the victim.

The shared stream is the aggravating factor. Because ids and secrets come from the same `Math.random()` sequence, an id that the API returns openly leaks information about a secret that the API never returns.

## Proposed fix

The proposed fix routes character selection through a cryptographically secure source while keeping the public behavior of `randomId()` unchanged. It would replace `randomChar()` in `libs/juava/src/id.ts` with the following:

```js
/**
 * Fills `size` bytes from a cryptographically secure source.
 *
 * `Math.random()` is deliberately NOT used here: `randomId()` mints bearer
 * credentials (API key secrets, CLI keys, workspace invitation tokens), and
 * V8's `Math.random()` is a PRNG whose internal state is recoverable from a
 * few observed outputs - letting an attacker predict subsequent tokens.
 *
 * Prefers the Web Crypto API (available in browsers and Node >= 19, and this
 * module is bundled into the browser console) and falls back to Node's
 * `crypto.randomBytes` for older/edge runtimes.
 */
function secureRandomBytes(size: number): Uint8Array {
  const webcrypto = typeof globalThis !== "undefined" ? (globalThis as any).crypto : undefined;
  if (webcrypto && typeof webcrypto.getRandomValues === "function") {
    return webcrypto.getRandomValues(new Uint8Array(size));
  }
  // eslint-disable-next-line @typescript-eslint/no-var-requires
  return require("crypto").randomBytes(size);
}

/**
 * Uniform integer in [0, max) using rejection sampling to avoid the modulo
 * bias that `byte % max` would introduce when `max` does not divide 256.
 */
function secureRandomInt(max: number): number {
  const limit = 256 - (256 % max); // largest multiple of `max` that is <= 256
  while (true) {
    const byte = secureRandomBytes(1)[0];
    if (byte < limit) {
      return byte % max;
    }
  }
}

function randomChar(noDigits?: boolean) {
  const chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
  while (true) {
    const index = secureRandomInt(chars.length);
    if (!noDigits || index > 9) {
      return chars[index];
    }
  }
}
```

The proposed fix rests on three decisions.

The source is the Web Crypto API first, with a Node fallback. `juava` ships raw TypeScript (`main` is `src/index.ts`), so a consumer's bundler transpiles `randomId()` into both Node services and the browser console. `globalThis.crypto.getRandomValues` exists in browsers and in Node 19 and later. The `require("crypto").randomBytes` branch covers runtimes without the Web Crypto global. A Node-only API such as `crypto.randomInt` would risk the browser build.

The mapping from bytes to characters uses rejection sampling. The alphabet holds 62 characters, and 62 does not divide 256. A direct `byte % 62` would map 256 byte values onto 62 characters unevenly, over-representing the first eight characters. `secureRandomInt` discards bytes at or above 248, the largest multiple of 62 that is at most 256, so every returned character is equally likely. Biased token characters shrink the effective keyspace without any visible error.

The public contract stays unchanged. `randomId()` keeps its signature, output length, alphabet, optional prefix, and the rule that the first character is never a digit, so no call site needs changing.

## Proposed tests

The proposed regression tests belong in `libs/juava/__tests__/id.test.ts` and run under Vitest with `pnpm --filter @jitsu/juava test`. They should assert both correctness and the security property.

Correctness tests should confirm output length, alphabet membership, the leading-non-digit rule, and absence of collisions across 5000 draws.

Two tests should guard against a regression to a weak source. The first spies on `Math.random` and asserts that `randomId()` never calls it:

```js
test("does not use Math.random()", () => {
  const spy = vi.spyOn(Math, "random");
  randomId(32);
  randomId({ digits: 16, prefix: "key" });
  expect(spy).not.toHaveBeenCalled();
});
```

The second pins `Math.random` to a constant and asserts that output stays varied. A generator that depends on `Math.random()` collapses to a single repeated character under this condition:

```js
test("output is independent of Math.random() (stays random when it is pinned)", () => {
  vi.spyOn(Math, "random").mockReturnValue(0);
  const id = randomId(64);
  const distinct = new Set(id.split(""));
  expect(distinct.size).toBeGreaterThan(1);
});
```

A distribution test concatenates many identifiers and asserts that the output covers at least 50 of the 62 alphabet characters. This test catches a broken byte-to-character mapping, such as faulty rejection sampling, that would collapse output onto a narrow subset.

After applying the proposed fix, the full `juava` suite should pass under `pnpm --filter @jitsu/juava test`, and the package should type-check with `tsc --noEmit`.

## Recommended follow-up

Applying the proposed fix secures newly generated credentials. It does not change credentials already minted by the current generator, which remain predictable to an attacker who captured earlier outputs from the same process. After the fix lands, rotating long-lived API keys, CLI keys, and outstanding workspace invitation tokens removes that residual exposure. This report does not cover the rotation procedure.

ciechanowiec:
Fixes #1356.

### DIFF
--- a/libs/juava/__tests__/id.test.ts
+++ b/libs/juava/__tests__/id.test.ts
@@ -1,4 +1,11 @@
+import { afterEach, expect, test, vi } from "vitest";
 import { randomId } from "../src";
+
+const ALPHABET = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
 
 test("id test", () => {
   const id1 = randomId();
@@ -14,4 +21,63 @@ test("id test", () => {
   expect(id3.length).toBe(10);
   expect(id4.length).toBe("test_".length + 10);
   expect(id4.startsWith("test_")).toBe(true);
+});
+
+test("only emits characters from the expected alphabet", () => {
+  for (let i = 0; i < 200; i++) {
+    const id = randomId(32);
+    for (const ch of id) {
+      expect(ALPHABET).toContain(ch);
+    }
+  }
+});
+
+test("first character is never a digit (so ids are valid identifiers)", () => {
+  for (let i = 0; i < 500; i++) {
+    const id = randomId(16);
+    expect(/^[0-9]/.test(id)).toBe(false);
+  }
+});
+
+// --- Security regression tests (CWE-338: weak randomness for secrets) ---
+//
+// randomId() mints bearer credentials (API key secrets, CLI keys, invitation
+// tokens). It MUST draw from a CSPRNG, never Math.random(), whose state is
+// recoverable from observed output, which would let an attacker predict tokens.
+
+test("does not use Math.random()", () => {
+  const spy = vi.spyOn(Math, "random");
+  randomId(32);
+  randomId({ digits: 16, prefix: "key" });
+  expect(spy).not.toHaveBeenCalled();
+});
+
+test("output is independent of Math.random() (stays random when it is pinned)", () => {
+  // Pin Math.random to a constant. A weak implementation would now emit a
+  // constant character; a CSPRNG-backed one keeps producing varied output.
+  vi.spyOn(Math, "random").mockReturnValue(0);
+  const id = randomId(64);
+  const distinct = new Set(id.split(""));
+  expect(distinct.size).toBeGreaterThan(1);
+});
+
+test("generates unique ids with no collisions across many draws", () => {
+  const seen = new Set<string>();
+  const n = 5000;
+  for (let i = 0; i < n; i++) {
+    seen.add(randomId(24));
+  }
+  expect(seen.size).toBe(n);
+});
+
+test("character distribution is broad (no modulo collapse / bias smoke test)", () => {
+  // Concatenate many ids and confirm the generator covers most of the
+  // 62-character alphabet — a buggy CSPRNG mapping (e.g. broken rejection
+  // sampling) would collapse onto a narrow subset.
+  let blob = "";
+  for (let i = 0; i < 200; i++) {
+    blob += randomId(32);
+  }
+  const distinct = new Set(blob.split(""));
+  expect(distinct.size).toBeGreaterThanOrEqual(50);
 });
--- a/libs/juava/__tests__/id.test.ts
+++ b/libs/juava/__tests__/id.test.ts
@@ -1,8 +1,6 @@
 import { afterEach, expect, test, vi } from "vitest";
 import { randomId } from "../src";
 
-const ALPHABET = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
-
 afterEach(() => {
   vi.restoreAllMocks();
 });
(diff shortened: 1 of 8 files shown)


vklimontovich:
Thanks for this, good catch. I pushed a commit on top (b465217ac) with a few changes.

Made the CSPRNG opt-in via a `strongRandom` flag. randomId is used a lot for plain ids where Math.random is fine, and the crypto path is ~36x slower per byte. So default stays Math.random, and you pass `randomId({ digits: 32, strongRandom: true })` for secrets.

Then I turned it on at the places that actually mint credentials: api/cli key plaintext, invitation tokens, the CSRF token, the generated password. The id fields next to them are just identifiers, left those alone.

Also dropped the require("crypto") fallback. globalThis.crypto is there in browsers, Node >=19 and edge, so it never ran, and a static require of a node builtin breaks edge/browser bundles.

And trimmed the tests down to the two that matter (strongRandom never calls Math.random, and stays random when Math.random is pinned).

Core fix is yours, thanks for digging into it. Let me know if you'd rather do it differently.

ciechanowiec:
@vklimontovich , works for me. Thank you for the update.

PR #1389, opened 2026-07-08
add internal slack notification for first Destination

that70schris:
adds slack notification for first destination and connector

### DIFF
--- a/webapps/console/lib/server/config-objects-service.ts
+++ b/webapps/console/lib/server/config-objects-service.ts
@@ -214,9 +214,32 @@ export class ConfigObjectsService {
     delete object.id;
     delete object.workspaceId;
     delete object.cloneId;
+
     const created = await this.prisma.configurationObject.create({
       data: { id, workspaceId, config: object, type },
     });
+
+    if (["destination", "service"].includes(type)) {
+      const count = await this.prisma.configurationObject.count({
+        where: { workspaceId, type: type, deleted: false },
+      });
+
+      if (count === 1) {
+        withProductAnalytics(
+          p =>
+            p.track("workspace_activated", {
+              ...objectAnalyticsProps(type, id, object),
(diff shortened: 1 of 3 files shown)


webapps/console/lib/server/serverEnv.ts
```diff
@@ -374,6 +374,7 @@ const ServerEnvSchema = ClientEnvSchema.extend({
 
   // Slack webhook URL for notifications
   SLACK_WEBHOOK_URL: z.string().optional(),
+  SLACK_WEBHOOK_URL_ACIVATIONS: z.string().optional(),
```
that70schris: @priyank-insurgrid @absorbb @vklimontovich not sure where this needs to be set

vklimontovich (changes requested):
Let's not expose a internal ops in the public repo. The best way to do this would be sending events to billing server, and billing server should send all the internal messages

webapps/console/lib/server/serverEnv.ts
vklimontovich: We should set it in the deployment repo, but also let's not post secret URLs to the public repo and let's rotate this secret

that70schris: it has been rotated

webapps/console/lib/server/config-objects-service.ts
```diff
@@ -189,9 +190,28 @@ export class ConfigObjectsService {
...
       data: { id, workspaceId, config: object, type },
     });
+
+    if (["destination", "service"].includes(type)) {
+      const count = await this.prisma.configurationObject.count({
+        where: { workspaceId, type: type, deleted: false },
+      });
+
+      if (count === 1) {
+        sendActivationSlackMessage({
```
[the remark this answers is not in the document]
that70schris: this is expected

vklimontovich (dismissed):

PR #1416, opened 2026-07-21
fix(functions): insert underscore between a digit and a capital letter in idToSnakeCaseFast

sarmah-rup:
### What

`idToSnakeCaseFast` in `libs/functions/src/lib/strings.ts` does not insert an underscore between a digit and a following capital letter. An id like `field1Name` becomes `field1name` instead of `field1_name`.

### Why I hit this

I run events into a warehouse through Jitsu and noticed some of my columns came out mis-named. Properties like `field1Name`, `address2Line` and `plan9FromOuterSpace` landed as `field1name`, `address2line` and `plan9from_outer_space` rather than the snake_case I expected. That breaks downstream models that reference `field1_name`.

### Root cause

In the loop, `needUnderscore` is only turned on after a latin letter, so the digit to capital boundary is missed:

```ts
// we add underscore only between latin letters
needUnderscore = (c >= aCode && c <= zCode) || (c >= ACode && c <= ZCode);
```

The comment even says "only between latin letters", but the canonical reference implementation `idToSnakeCaseRegex` in `libs/core-functions-lib/src/functions/lib/strings.ts` uses a `(?<=[a-zA-Z0-9])` lookbehind, which includes digits. So the fast and regex versions disagree exactly on digit boundaries.

### Fix

Include the digit range in `needUnderscore`, reusing the `zeroCode` / `nineCode` constants already declared at the top of the file:

```ts
needUnderscore = (c >= aCode && c <= zCode) || (c >= ACode && c <= ZCode) || (c >= zeroCode && c <= nineCode);
```

### Tests

Added regression cases to `libs/core-functions-lib/__tests__/strings.test.ts` asserting the corrected output (`field1Name` becomes `field1_name`, `address2Line` becomes `address2_line`, `plan9FromOuterSpace` becomes `plan9_from_outer_space`) and that `idToSnakeCaseFast` now matches `idToSnakeCaseRegex`. The two new tests fail on `newjitsu` before the change and pass after. The full `core-functions-lib` suite is green at 42 tests.

### DIFF
--- a/libs/core-functions-lib/__tests__/strings.test.ts
+++ b/libs/core-functions-lib/__tests__/strings.test.ts
@@ -59,6 +59,13 @@ const dataExtra: Record<string, string> = {
   "$Camel##Case#": "$camel##case#",
 };
 
+const dataDigits: Record<string, string> = {
+  // underscore must be inserted between a digit and a following capital letter
+  field1Name: "field1_name",
+  address2Line: "address2_line",
+  plan9FromOuterSpace: "plan9_from_outer_space",
+};
+
 test("test idToSnakeCaseFast", async () => {
   for (const [value, expected] of Object.entries(data)) {
     const res = idToSnakeCaseFast(value);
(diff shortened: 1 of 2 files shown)


vklimontovich:
Hey @sarmah-rup, thank you for the PR! If we merge it, would it mean that columns containing digits will be renamed? May be too disruptive

PR #1441, opened 2026-08-04
JITSU-158: explicit typing + unsafe-any lint gate on config exports

absorbb:
Action items **2** and **3** of `JITSU-158` (2026-07-30 blank-options postmortem), plus the console half of the JITSU-139 export-resilience work.

> **Stacked on #1433** (`fix/jitsu-139-export-resilience`) — merge #1433 first, then this retargets to `newjitsu` automatically.
>
> **This PR now owns the whole console export change.** #1433 and this PR both rewrote `pages/api/admin/export/[name]/index.ts`, so the two export-resilience commits moved here and #1433 is now config-keeper / bulker / rotor only. Nothing was dropped — the two PRs together are byte-identical to what they contained before the split.

## Export resilience (moved from #1433)

One malformed entity can no longer poison a whole `/api/admin/export/*` export: each entity's JSON is built before anything is written, and malformed entities are skipped and logged with the `System error:` marker — including the `clickhouseSettings` null-deref that triggered JITSU-139. If an export still fails mid-stream the socket is destroyed, so consumers see an aborted transfer instead of valid-looking truncated JSON with status 200. The syncs export fails closed rather than emitting a partial set.

This is the same defence the typing work below approaches from the other side: resilience keeps a bad row from breaking the stream at runtime, the types keep the bad row from being written in the first place.

## Item 2 — break the silent-`any` (commit 1)

Every paginated `findMany` in `pages/api/admin/export/[name]/index.ts` now carries an explicit `Prisma.…GetPayload` annotation. The cursor back-edge (`cursor:` ← `lastId` ← `objects[last].id`) made inference self-referential, and the checker silently collapsed the result to `any` — which is how `{ data_ }` (for `{ data: data_ }`) compiled on July 30.

**Verified:** re-introducing the exact incident typo now fails with `TS2339: Property 'data_' does not exist on type '{ from: …; to: …; workspace: … }'`.

Along the way:
- All Prisma `Json` column reads (`link.data`, `object.config`, `pb.connectionOptions`, `intermediateStorageCredentials`) go through tolerant zod parsers — field-level `catch` drops only the offending field, non-object roots normalize to `{}`, so a junk row still can't fail the export (consistent with the per-entity resilience above).
- `$queryRaw` / billing `pgPool` queries get typed results; `rpc()` backup-connections result is validated as an array instead of trusted as `any`.
- `object-hash` / `stable-hash` imports pinned to explicit signatures (no shipped types / `exports` map without a `types` condition — their inferred type differs between tsc and the type-aware lint program).
- `getCoreDestinationTypeNonStrict` accepts `string | undefined` — call sites always could pass `undefined`; the signature just couldn't say so.

Behavior preserved, except three crash→skip fixes on malformed rows (domain config without string `name`, classic-mapping without string `value`, canceled subscription with null `period_end` — each previously threw mid-export).

## Item 3 — lint gate (commit 2)

`@typescript-eslint/no-unsafe-assignment` / `-member-access` / `-argument` as **errors** on `pages/api/admin/**`, type-aware via the project tsconfig. Runs inside `pnpm lint`, already part of the required **✨ Lint & Test** check — no CI changes needed.

The export endpoints pass clean. Eight admin files with pre-existing implicit-`any` debt (261 findings) carry a file-top exemption header scoped to exactly the rules they violate — the ratchet: remove the header when fixing a file, never add new ones.

## Verification
- `tsc --noEmit` — clean
- `eslint .` — 0 errors (2 pre-existing react-hooks warnings, untouched files)
- `vitest --project unit` — 30/30
- incident-typo repro — fails to compile (see above)


absorbb:
Added a third commit per discussion: connection options now parse through the destination type's own `connectionOptions` schema (`destinations.tsx`) with `.passthrough()`, generic tolerant parse only as the corner-case fallback (unknown type / non-conforming stored data, logged at warn).

Consequence (deliberate): absent fields materialize to console defaults — the export emits explicit `deduplicate: true, mode: batch, frequency: 60, …`, so consumers never re-default absent options themselves (`JITSU-136`, JITSU-158 item 8 console-side). A blank-`data` row — the exact 2026-07-30 failure shape — now exports safe defaults instead of nothing. Stored values win over defaults; unknown keys flow through.

Heads-up for rollout: one-time `optionsHash` churn across all connections, and connections that never persisted `frequency` move from bulker's fast absent-default (~1 min) to the 60 min the UI has always claimed.

absorbb:
Back-compat amendment: `frequency` is now **excluded** from default materialization — it exports only when actually stored (explicit `null` included), so connections that never persisted a frequency keep bulker's current cadence. All other defaults (`deduplicate`, `mode`, `primaryKey`, `dataLayout`, `batchSize`, …) still materialize.

absorbb:
One more hardening after checking prod data: `.passthrough()` only applies to the object it's called on, and 3 live connections carry an `enabled` flag *inside* `functions[]` entries that the nested schema would have stripped. Added `deepPassthrough` — rebuilds the per-type schema with passthrough at every object level, so stored fields survive at any depth. Defaults and validation semantics unchanged (tested: nested extras kept, blank-{} still materializes defaults, invalid entries still fall back).

PR #1449, opened 2026-08-07
feat(destinations): Google Ads conversions via Data Manager API and legacy Ads API

vklimontovich:
Adds a `google-ads` cloud destination that forwards Jitsu events as Google Ads conversions.
Jitsu has had no server-side Google Ads path at all — unlike Meta, where `facebook-conversions`
pairs with the pixel, the tag workaround physically cannot do offline or enhanced upload.

## Why two APIs

Google **closed `UploadClickConversions` to new adopters on 2026-06-15** — new callers get
`CUSTOMER_NOT_ALLOWLISTED_FOR_THIS_FEATURE`, and it separately requires an approved developer
token. The designated replacement is the **Data Manager API** (`datamanager.googleapis.com`):
no developer token, one OAuth scope.

So the connector defaults to Data Manager and keeps the legacy Ads API behind an `api` selector
for accounts already allowlisted. One canonical `Conversion` is built from the event, then handed
to one of two serializers — the APIs carry the same data under different names, and disagree on
almost every field (`currency`/`currencyCode`, `transactionId`/`orderId`, RFC3339 vs
`yyyy-mm-dd hh:mm:ss+hh:mm`, `CONSENT_GRANTED`/`GRANTED`, nested vs top-level click ids).

## What "Google conversion API" covers

| Feature | Legacy Ads API | Data Manager | Status |
|---|---|---|---|
| Offline click conversions (gclid) | `uploadClickConversions` | `events:ingest` | ✅ |
| Enhanced Conversions for **Leads** | same call + `userIdentifiers` | same call + `userData` | ✅ |
| Enhanced Conversions for **Web** | **`uploadConversionAdjustments`** (`ENHANCEMENT`) | `events:ingest` + `transactionId` | ✅ |
| Floodlight / CM360 | — | — | out of scope |

Leads and offline are genuinely the same call. **Web is a different endpoint on the legacy API** —
an *adjustment* that enriches a conversion the on-page tag already recorded, joined on order ID.
Selected via `conversionType: conversion | enhancement`.

## gclid capture

Jitsu dropped `gclid` entirely — `context.clientIds` carried only fbc/fbp/GA4, and `parseUtms`
keeps only `utm_*`. Fixed in two layers:

1. **SDK** — `getClientIds()` now reads `gclid`/`gbraid`/`wbraid`/`dclid` from URL params and the
   `_gcl_aw`/`_gcl_dc` cookies, mirroring how `fbc`/`fbp` already work.
2. **Server** — the destination function runs for every event on the connection (the `events`
   filter is applied inside it), so it remembers any click id under `gads_click:<anonymousId>`
   for 90 days and replays it onto a conversion that fires later. Degrades to "no replay" when no
   Mongo/Redis is configured; never throws.

Reads resolve `context.clientIds` → page URL/search (per field, filling gaps) → remembered value.
Layer 2 and the URL parse are what make this work for customers who haven't upgraded the SDK.

## Notable

- Both APIs **return HTTP 200 while rejecting rows** (`partialFailureError` / `fieldWarnings`).
  Checking `response.ok` alone silently drops conversions, so both bodies are inspected.
- `transactionId` falls back to `messageId`, which is fine for deduping a new conversion but can
  never match a tag-recorded one — so `orderId` (real order ID only) is tracked separately and
  enhancement mode requires it, rather than sending a request Google accepts and matches to nothing.
- The developer token (legacy API only) identifies the calling application rather than the
  advertiser, so it falls back to `GOOGLE_ADS_DEVELOPER_TOKEN` instead of being asked of each user.
- `facebookHash`/`sanitizePhone` hoisted into `functions/lib/pii.ts` and shared. Google's gmail
  dot/plus normalization is Google-specific and deliberately kept out of the shared helper.

## Testing

`pnpm --filter @jitsu/destination-functions test` — 14 tests covering both serializers, the
consent-enum and timestamp divergences, click-id replay and store degradation, the HTTP-200
rejection paths, and the enhancement-mode skip rules. Plus an env-gated live test
(`TEST_GOOGLE_ADS_DESTINATION`) that runs with `validateOnly: true`, so it validates against a real
account without recording conversions — **not yet run**, since it needs live credentials.


vklimontovich:
Both review findings addressed in ee35286d8.

**Click-id fan-out** — you're right, and the gap was worse than just `userId`: a `gclid` first seen while the user was anonymous became unreachable once they logged in on another device. Read path now backfills the remembered set onto the user's other identifiers, so it converges instead of staying pinned to whichever key happened to see the click.

**IP-only matching** — checked this against Google's docs and it splits by API. Data Manager explicitly lists `eventDeviceInfo.ipAddress` among the identifiers it will match on ("set at least one of: adIdentifiers…, session attributes, userData, or eventDeviceInfo.ipAddress"), so IP-only is legitimate there. The legacy API only takes `userIpAddress" alongside a click id or user identifiers, so those rows were guaranteed rejections exactly as you suspected. The guard is now per-API rather than dropped or kept wholesale.

Two tests added for these.

vklimontovich:
Right again — fixed in 720118fa8.

`dclid` exists on Data Manager's `AdIdentifiers` but has no field on the legacy `ClickConversion`, so a dclid-only event passed `hasClickId` and then serialized to a row with no identifier at all. Guaranteed rejection, exactly as you describe.

Rather than special-casing `dclid`, the gate now counts only the click ids the selected API can actually serialize (`LEGACY_CLICK_ID_NAMES` vs the full set). That keeps the capture side unchanged — a dclid is still captured and remembered, since switching the destination to Data Manager makes it useful — while stopping it from being mistaken for a match signal in legacy mode.

Test added covering both directions.

PR #1452, opened 2026-08-10
fix(intercom): send last_seen_at as unix seconds when upserting contacts

tevaum:
## What
- Send `last_seen_at` as an integer unix timestamp in `createOrUpdateContact`, matching the event and `updateLastSeenOnEveryEvent` paths in the same file
- Add mocked create/update tests asserting the payload Intercom receives

## Why
Intercom's contacts API rejects the ISO-8601 string with `400 parameter_invalid`
("'last_seen_at' must be a Unix timestamp in seconds (integer), got String"), so every
`identify` event carrying an email failed to create or update its contact. The payload is
used for both `POST /contacts` and `PUT /contacts/:id`, so contact sync fails in both
branches. Observed on 2.8.5 and reproduced on current `newjitsu`.

### DIFF
--- a/libs/destination-functions/__tests__/intercom-contact.test.ts
+++ b/libs/destination-functions/__tests__/intercom-contact.test.ts
@@ -1,0 +1,65 @@
+import { afterAll, afterEach, beforeAll, expect, test } from "vitest";
+import { http, HttpResponse } from "msw";
+import { setupServer } from "msw/node";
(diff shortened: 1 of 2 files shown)


tevaum:
Let me know if a GH issue is also needed. I can create one with details and logs of the specific problem I'm having on my instance.

PR #1461, opened 2026-08-14
JITSU-164: shard the bulker test suite by destination and isolate runs

absorbb:
`JITSU-164`

The `bulker-test` job took ~13.5 min, and two workflow runs in flight at once corrupted each
other's state in the shared cloud warehouses. This is parts **C** (isolation) and **B**
(sharding) of that ticket, plus two things the work turned up along the way.

## C — one schema per run

Table names come from test names and the schema (dataset, on BigQuery) came fixed from the
destination secret, so concurrent runs created, asserted on and dropped the *same* objects.

`BULKER_TEST_RUN_ID` now rewrites the destination config's schema field — `bqDataset` for
BigQuery, `defaultSchema` for Snowflake / Redshift / MotherDuck — to `bulker_ci_<run id>`, and
the namespaces tests ask for explicitly are prefixed to match. Isolation at the schema level
rather than a table-name prefix keeps `TestNaming`'s exact-name assertions working unchanged.

The run's schemas are created once up front, over a connection that still uses the
destination's original schema. That turned out to be load-bearing:

- **Snowflake** pins the schema at *connection* time and returns `390201 (08004)`, so
  `create_bulker` failed before `InitDatabase` could create anything.
- **BigQuery** does create on demand, but with a brand-new dataset every run the first wave of
  parallel tests raced to create it — `409 Already Exists` and `403 Exceeded rate limits: too
  many dataset metadata update operations`. Before isolation the dataset always existed, so
  `Create` was never called.

Cleanup drops the run's schemas on `always()`, so it also fires on cancellation — which, with
`cancel-in-progress` already on this workflow, is what happens on every re-push.
`ownedByTestRun` makes it structurally unable to drop a schema the run doesn't own.

Leaving `BULKER_TEST_RUN_ID` unset (local runs) keeps the historical names.

## B — one shard per destination

`BULKER_TEST_CONFIGS` filters the destination list before anything is created, so a cloud shard
no longer starts the local containers it will never use. `bulker-test` becomes a matrix: `app`
(everything that isn't the sql package), `local` (the five containerised destinations), and one
shard each for bigquery / snowflake / redshift / motherduck. Each shard gets only the secret it
needs, and `fail-fast: false` keeps one broken warehouse from hiding the others.

The `app` shard's package list plus `./bulkerlib/implementations/sql` was diffed against the
old `./bulkerlib/...` expansion via `go list` — coverage is identical.

Sharding also broke the Go build cache: six jobs shared one `setup-go` key while building
different packages, so only the first to finish could save it (`Unable to reserve cache with
key …, another job may be creating this cache`) and the rest restored someone else's. Modules
and build output are now cached separately, the build cache keyed per shard and per run so it
stays fresh rather than being written once and going stale.

## The required status check

The `default-branch-protection` ruleset requires a check named exactly `🧪 Bulker Test`.
Sharding renamed it to `🧪 Bulker Test (app)` and friends, which would have left every PR
unmergeable. A `bulker-test-result` job carries that name and gates on the whole matrix, so the
ruleset needs no edit — and neither does adding a shard later.

## bulkerapp container sharing

Each of `bulkerapp`'s three integration tests started its own Postgres *and* Kafka: 126s of the
package's 296s was container startup, paid three times. They now share one pair, started
lazily and torn down in `TestMain`, so a run touching only the unit tests still pays nothing.
Kafka could never have been per-test anyway — its compose stack has a fixed id, fixed container
names and a fixed host port, and it tears down any existing stack on the way up.

The cost is that these tests are no longer isolated from one another. `TestEventsRetry`'s
tables are renamed accordingly: it reused `TestGoodAndBadStreams`' `good_batch`/`good_stream`
against the same destinations, which with a shared Postgres and Kafka would have counted its
rows and replayed its topics. It also stops Postgres mid-test, so it now restores it in
`t.Cleanup` — otherwise a failed assertion would leave the shared container down for
everything after it.

Running the three in parallel would be the bigger win, but it's out of reach here:
`appbase.InitAppConfig` configures apps through the **global viper singleton**, so two apps
can't even be constructed concurrently, and unpicking that touches the startup config of all
seven Go services.

## Two silent-failure fixes

- A destination whose credentials are missing was silently dropped from the run. A shard that
  names a destination it can't configure now fails loudly rather than running zero tests and
  reporting green.
- That surfaced one immediately: `BULKER_TEST_REDSHIFT_SERVERLESS` is wired into the workflow
  but has never resolved — runs initialise `[bigquery redshift_iam snowflake postgres mysql
  clickhouse clickhouse_cluster clickhouse_cluster_noshards duckdb]`, no serverless. The shards
  match what really runs; `redshift_serverless` (and the long-commented-out `redshift`) get a
  shard once their secrets exist.

## Result

All six shards green. No warehouse gates any other, and a re-push no longer corrupts a
concurrent run. Wall clock on the latest run: `redshift` 7m16s is the critical path, with
`local` 6m15s, `app` 5m13s, `bigquery` 2m51s, `snowflake` 2m31s, `motherduck` 1m41s — against
13m27s for the single job before.

## Follow-ups (not here)

- **Part A**: `t.Parallel()` on the per-destination subtests in `runTestConfig` plus a higher
  `-parallel`. Needs two guards — `timestamp.FreezeTime()` is process-global so the
  `mergewindow` cases must stay serial, and `TestReconnect` stops/starts the shared containers.
- `TestMergeWindow` already calls `t.Parallel()` at the top level today, so its global time
  freeze already leaks into concurrently running tests. Latent flake, worth fixing on its own.
- Per-instance viper in `appbase`, which is what would unlock parallel `bulkerapp` tests.
- A janitor for schemas leaked by a runner that dies outright, past the cancellation grace
  period, which `always()` can't cover.


absorbb:
### First CI run: 4 of 6 shards green, 2 real bugs found and fixed

[Run 31804648106](https://github.com/jitsucom/jitsu/actions/runs/31804648106) — the shards do run fully in parallel, and per-run schema isolation works end-to-end on Redshift, MotherDuck and the local containers.

| shard | result | duration |
|---|---|---|
| app | ✅ | 12m06s |
| redshift | ✅ | 7m21s |
| local | ✅ | 6m43s |
| bigquery | ❌ | 5m02s |
| motherduck | ✅ | 4m02s |
| snowflake | ❌ | 3m11s |

Both failures came from the same wrong assumption — that pointing a config at a fresh schema is enough because adapters create schemas on demand:

- **Snowflake** pins the schema at *connection* time and returns `390201 (08004): The requested schema does not exist or not authorized`, so `create_bulker` failed before `InitDatabase` could create anything. All 168 leaves failed this way.
- **BigQuery** does create on demand, but the dataset is new every run now, so the first wave of parallel tests raced to create the same one: 9× `409 Already Exists` and 2× `403 Exceeded rate limits: too many dataset metadata update operations`. 11 failures at `init_database`, and the other 2 were later stages of multi-stage tests whose earlier stage had failed. Before isolation the dataset always existed, so nothing ever called `Create`.

Fixed by creating the run's schemas once, up front, over a connection that still uses the destination's original schema. Cleanup now uses that connection too, so dropping no longer depends on the schema being there.

### On the wall-clock

The destination suite is no longer the gate — Redshift went from holding up everything to 7m21s in its own lane, and a change now gets local feedback in under 7 minutes. But **the `app` shard is now the critical path at 12m06s**, so end-to-end this run was 13m27s → 12m06s.

That shard is `bulkerapp/app` (296s) + file_storage (77s) + kafkabase (40s) against a 617s `go test` span, i.e. a large chunk is compilation — and this was the branch's first run, so every shard had a cold Go build cache. Worth a follow-up either way: splitting the destinations out has exposed `bulkerapp` as its own ~12 minute problem, which was previously hidden behind Redshift.

PR #1508, opened 2026-09-10
JITSU-163: zstd for S3/GCS destinations and internal batch paths

sahiltyagi-jitsu:
Adds `FileCompressionZSTD` end to end for S3/GCS destinations and internal batch paths. **Merging this changes nothing at runtime** — every new zstd write is behind config defaulting to gzip, and the `jitsu-cloud-billing` change that flips the event-archive connection lands separately, once every reader is deployed.

Stacked on #1506. That must merge first.

**Updated after review:** the job-payload codec is removed — `admin` compresses job config with gzip as before and `reprocessing-worker` assumes gzip, per @absorbb. `admin` is now byte-identical to its pre-change state apart from the `failover_reprocessor` filters. Re-verified end to end afterwards on both backends.

## The framing that shaped the diff

The ticket is organised by *where compression happens*. The risk is in *where filenames are built and matched*, and that list is longer.

Every reader here picks its decompressor from a file name. So a missed compression branch fails loudly — no zstd gets written. A missed **naming** branch fails silently: zstd bytes under a gzip-shaped name are misread, skipped by discovery, or never cleaned up. Suffixes therefore live in one place — `jitsubase/compression`, the module every reader and writer already imports.

## Sites the ticket doesn't list

Found by grepping the invariant rather than following the ticket's list. Ildar already asked for the `failover_reprocessor` filters in this PR; these are additional:

- **`implementations/file.go` `AddFileExtension`** — the `switch` only handled GZIP, so a zstd object would be named `.ndjson` with zstd bytes inside. Live on the S3/GCS write path via `file_storage/abstract.go`. The worst of these.
- **`s3.go` and `google_cloud_storage.go`** set `Content-Type`; zstd would have fallen through to `application/x-ndjson`. The ticket names these files but frames them only as reading `config.Compression`.
- **`kafkabase/failover_logger.go`** globs `*.ndjson` / `*.ndjson.gz` when cleaning up, so `.zst` files would never be deleted — local disk grows unbounded. Different and worse than being skipped.
- `FileExtension()` in the three marshallers, and the hardcoded `".gz"` on failover rotate.
- `blockStorageSettings.compression` in the console schema — zstd wasn't selectable at all. Default stays gzip, per the ticket.

## Departures from the ticket

**1. Encoder pools, not one encoder per destination.** A `Destination` is shared across one topic consumer per table × mode, so several batches run against it concurrently, and zstd's streaming writer isn't concurrency-safe.

Not a preference — demonstrated. `TestZstdMarshallersAreSafeUnderConcurrency` runs 16 goroutines through the pool and passes under `-race`. Replacing the pool with one shared encoder, which is what the ticket says literally, produces `WARNING: DATA RACE` six times.

**2. A shared suffix package.** Not in the ticket. Adds no dependency edges — all four modules already import `jitsubase`. Holds only strings and matching; codec wiring stays with the callers.

**3. A small refactor.** The decompressor selection was inline in a large function in `reprocessing-worker`, which is why it had no coverage. Now `decompressorFor(path, r)`.

**4. `file_storage/abstract.go` deliberately keeps stdlib `compress/gzip`** at its default level. It was never part of the klauspost swap, and routing it through the shared helper would change existing gzip output as a side effect of adding zstd. **This asymmetry is intentional and easy to "tidy" into a regression.**

## Verified in helm dev

Binary provenance checked first — the running bulker carries this branch's symbols.

| Destination | Object | Content-Type |
|---|---|---|
| S3 zstd | `zstd-ndjson/events_….ndjson.zst` | `application/zstd` |
| S3 gzip (ndjson + csv) | `events_….ndjson.gz` / `.csv.gz` | — |
| GCS zstd | `zstd-ndjson/events_….ndjson.zst` | `application/zstd` |
| GCS gzip | `gzip-ndjson/events_….ndjson.gz` | `application/gzip` |

~12k events through `POST /api/s/s2s/batch`. The S3 zstd object is read by the `zstd` CLI and Python's stdlib; `file` reports Zstandard; 4000 valid JSON records.

**The result that matters most:** on both backends the gzip and zstd objects decompress to **the same 4000 events** — equivalent record for record once field order is normalised, and the same 1,934,847 bytes in length. Evidence the gzip path is untouched.

They are deliberately *not* compared byte for byte: `Object` is built from a Go map, so field order is randomised between runs and byte-equality would fail on correct code. An earlier version of this section claimed byte-identity, which was inferred from equal length rather than measured.

GCS was tested against `fsouza/fake-gcs-server` via `STORAGE_EMULATOR_HOST`, not real GCS.

## Backward compatibility

The reader and filter changes are **not** config-gated — they take effect on deploy — so each was checked as equivalent rather than assumed safe:

- the discovery matcher is a **strict superset**: only `.ndjson.zst` newly matches, checked against the old logic including pathological double extensions
- the cleanup globs add a pattern that matches nothing today
- `decompressorFor` is identical for `.gz` and uncompressed
- `CompressionExtension(GZIP)` is still `.gz`; `CompressionContentType(GZIP)` is still `application/gzip`

## Replay

`TestReprocessIsIdenticalAcrossCodecs` drives the **real** reprocess function over real files, once per codec, requiring identical events, counts and `connection_ids`.

Mutation-checked. With the zstd case removed:

```
TotalLines = 1, want 4 - the file was probably not decompressed
...ndjson.zst replayed [], want [e1 e2 e3 e4]
```

Zero events replayed, one parse error as the only signal — the exact failure the prerequisite exists to prevent.

## Memory

The ticket flags +18–21 MB RSS per concurrent encoder; pooling was added on top, so it needed numbers.

~20.6 MB per concurrent encoder, flat from 1 to 64. **The pool does not retain them** — heap returns to baseline after two GCs (1154 MB → 660 MB after one GC, 0.4 MB after two). So the cost is peak, not a leak: **peak RSS scales linearly with concurrent zstd batches** — 64 at once is over 1 GB. `zstd.WithWindowSize` is the lever if it bites.

## What isn't covered

- **Real S3/GCS** — minio and an emulator stand in.
- **Streaming reads.** The replay test reads local files; production streams from object storage. Same `decompressorFor` either way.
- `mixpanel.go` needs a real token.

**Recommendation:** merge freely — it's a runtime no-op. Before flipping the archive to zstd, read one real `.zst` back through `reprocessing-worker` in the real environment. The config gate exists precisely so that can happen before any producer emits zstd.

## Notes

- klauspost stays at **v1.19.1**, not the benchmarked v1.19.2 — proven byte-identical at levels 4 and 6 by SHA-256, so the ticket's numbers transfer.
- One new env var: `FAILOVER_LOGGER_COMPRESSION_CODEC`, defaulting to `gzip`.
- `operator.go` stays on gzip per the ticket. Worth noting it builds `.json.gz` keys in **three** places (977, 1108, 1220), not one, if that's ever revisited.


absorbb:
It appeared quite heavier than i expected. Lets remove small-win changes that brought a lot of extra code:
**admin** - no need to add `JobPayloadCompressionCodec` - it's ok to compress job config with gzip
**reprocessing-worker**  - no need for `decompressBlob` - lets assume gzip as usual

absorbb (dismissed):

absorbb (approved):

PR #1528, opened 2026-09-22
fix(operator): give functions-server a StartupProbe (JITSU-232)

sahiltyagi-jitsu:
Fixes JITSU-232. Fix for the `free-3-5-fs` CrashLoopBackOff page on 21 Sep.

## Problem

The functions-server container has a livenessProbe and no startupProbe:

```
livenessProbe: initialDelay 10s, period 10s, failureThreshold 5, timeout 5s
startupProbe:  NONE
```

That is a ~60s deadline from container start, and liveness begins counting before the server is up. Measured container-start → ContainersReady across every functions-server deployment in `newjitsu`:

```
deployment                      n   median    max
free-3-5-fs                     9     53s     61s   <- over the deadline
free-4-5-fs                     2     17s     17s
every dedicated -fs            2+    7-11s   8-13s
```

`free-3-5` is already over budget. On 21 Sep a config change rolled the deployment and 13 of 29 pods restarted, 75 restarts total. The crashed pod's last log was 14:03:21 and it died at 14:04:38 — 77 seconds, still initialising, never having served `/health`.

Agreed with @Ildar as the general fix: *"adding startupProbe is good general solution for all problems of that kind"*.

## What changed

**`operator.go`** — the functions-server container gets a StartupProbe on `/health`: initialDelay 5, period 5, timeout 3, failureThreshold 60, a **300s budget**. Liveness and readiness are suspended until it passes. `mongobetween` is left alone; it is a TCP proxy with a trivial start.

**`config.go`** — a `podSpecVersion` constant is mixed into `CalculateOperatorConfigHash()`.

That second change is the non-obvious half. The reconcile loop only updates an existing deployment when `ShutdownAt` / `ConfigHash` / `OperatorConfigHash` / `FunctionsClass` change (`operator.go:296-311`). Changing the pod spec in code is therefore not enough on its own — existing deployments would keep the old spec until something unrelated changed, which on a quiet shard could be days. Bumping `podSpecVersion` forces exactly one rollout on deploy, and future pod-spec changes need the same bump.

## Testing

`operator/startup_probe_test.go` — first tests in this package. Five: the probe exists, its budget clears the worst observed startup with margin, liveness is undisturbed, `podSpecVersion` changes the operator config hash, and the hash is stable for identical config. Each was checked to fail when the probe is removed or its threshold lowered.

Behavioural validation on minikube — two identical deployments, a container that sleeps 90s before serving `/health` (longer than the old ~60s budget, inside the new one), the only difference being the StartupProbe:

```
probe-old   0/1 Running, restarts=3 and climbing, never Ready
            "Container slow failed liveness probe, will be restarted" x3
probe-new   1/1 Running, restarts=0, start->ready 106s
```

The old configuration reproduces the incident exactly; the new one starts cleanly.

## Scope and what this does not fix

This stops the crashing. It does not make startup fast — a 53s start still means slow rollouts and slow recovery, and ~30s of it happens in a phase that emits no logs at all. Those are separate tickets (T2 churned-workspace config on the shared shard, T3 instrumenting the function-loading path), deliberately not widened into here.

The minikube run validates probe *semantics*; the Go tests cover the operator generating the manifest. Both halves are covered, by different tests rather than one end-to-end run.


bulker/operator/config.go
```diff
@@ -96,11 +96,34 @@ func (c *Config) PostInit(settings *appbase.AppSettings) error {
...
+// hash. Nothing should assign to it outside tests.
+var podSpecVersion = "v2-startupprobe"
+
 // CalculateOperatorConfigHash calculates a hash of Config fields that affect deployments.
 // When this hash changes, deployments should be updated to reflect the new configuration.
 func (c *Config) CalculateOperatorConfigHash() string {
 	h := sha256.New()
 
+	// Not a Config field: forces a roll when the generated pod spec changes.
+	h.Write([]byte(podSpecVersion))
```
absorbb: let's not do it.
I want to avoid massive rollout of all functions-servers (including paid) one.
free functions-servers get updated quite often - so they all will catch startup probe settings eventually

